PR #11, opened 2019-06-24
Fixing parsing

daniellasry:
Parsing works!

Would be nice to refactor the Ganglion::read_thread() function a little bit. But it works!


Andrey1994:
Cool!
I will add mock for ganglionlib.dll update bindings and refactor some code

PR #33, opened 2020-04-20
Get raw, unfiltered data from Brainbit

mesca:
The current implementation streams filtered data from Brainbit.

Before the fix (2-30 Hz bandpass and 50 Hz bandstop) :
Before

After the fix:
After

### DIFF
--- a/src/board_controller/neuromd/brainbit.cpp
+++ b/src/board_controller/neuromd/brainbit.cpp
@@ -170,7 +170,7 @@ int BrainBit::prepare_session ()
                 {
 
                     signal_t3 =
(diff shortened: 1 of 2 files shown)


Andrey1994:
Nice, how did you figure out it?

mesca:
Just noticed this bandstop in my data and hoped that the data was filtered in the SDK and not on board. The C function is not documented, but there is a hint in the .NET examples.

Andrey1994:
cool, and guys from brainbit confirmed that it will not be removed in future versions and correct https://gitlab.com/brainbit-inc/brainbit-sdk/-/issues/24

PR #40, opened 2020-05-07
parsing impedance values for ganglion

Andrey1994:

### DIFF
--- a/cpp-package/src/board_shim.cpp
+++ b/cpp-package/src/board_shim.cpp
@@ -435,3 +435,14 @@ int *BoardShim::get_temperature_channels (int board_id, int *len)
     }
     return temperature_channels;
 }
+
+int *BoardShim::get_resistance_channels (int board_id, int *len)
+{
+    int *resistance_channels = new int[MAX_CHANNELS];
+    int res = ::get_eeg_channels (board_id, resistance_channels, len);
+    if (res != STATUS_OK)
+    {
+        throw BrainFlowException ("failed get board info", res);
+    }
+    return resistance_channels;
+}
--- a/csharp-package/brainflow/brainflow/board_shim.cs
+++ b/csharp-package/brainflow/brainflow/board_shim.cs
@@ -398,6 +398,29 @@ namespace brainflow
         }
 
         /// <summary>
+        /// get resistance channels for this board
+        /// </summary>
+        /// <param name="board_id"></param>
+        /// <returns>array of row nums</returns>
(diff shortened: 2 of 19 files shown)


src/board_controller/openbci/ganglion.cpp
```diff
@@ -372,36 +434,62 @@ int Ganglion::config_board (char *config)
...
+        {
+            start_command = "z";
+            stop_command = "Z";
+        }
+        else
+        {
+            if ((strlen (config)) && (config[0] == 'Z'))
+            {
+                start_command = "b";
+                stop_command = "s";
```
daniellasry, line 451: I suggest declaring "b" "s" "z" and "Z" as named constants somewhere. And make a couple functions to switch the start/stop commands from standart to impedance (because i see it done in two places)

cpp-package/src/board_shim.cpp
```diff
@@ -435,3 +435,14 @@ int *BoardShim::get_temperature_channels (int board_id, int *len)
     }
     return temperature_channels;
 }
+
+int *BoardShim::get_resistance_channels (int board_id, int *len)
+{
+    int *resistance_channels = new int[MAX_CHANNELS];
```
daniellasry, line 441: I didn't catch this easrlyer butr what are you using "new" to initialize arrays of a constant size?

daniellasry: I guess because you are returning the pointer

csharp-package/brainflow/brainflow/board_shim.cs
```diff
@@ -397,6 +397,29 @@ public static int[] get_temperature_channels (int board_id)
...
+        /// <summary>
+        /// get resistance channels for this board
+        /// </summary>
+        /// <param name="board_id"></param>
+        /// <returns>array of row nums</returns>
+        /// <exception cref="BrainFlowException">If this board has no such data exit code is UNSUPPORTED_BOARD_ERROR</exception>
+        public static int[] get_resistance_channels(int board_id)
+        {
+            int[] len = new int[1];
+            int[] channels = new int[512];
```
daniellasry, line 409: What about here? Allocating on the heap but these arrays are not being returned?

daniellasry: Nevermind, this is c#. I'm not 100% awake yet 😄

cpp-package/src/board_shim.cpp
Andrey1994: exactly you can not return pointer to local variable

PR #41, opened 2020-05-08
send default settings in prepare_session

Andrey1994:

### DIFF
--- a/src/board_controller/openbci/cyton.cpp
+++ b/src/board_controller/openbci/cyton.cpp
@@ -8,6 +8,22 @@
 #define END_BYTE_ANALOG 0xC1
 #define END_BYTE_MAX 0xC6
 
+
+// load default settings for cyton boards
+int Cyton::prepare_session ()
+{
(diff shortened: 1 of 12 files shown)


src/board_controller/openbci/cyton.cpp
```diff
@@ -8,6 +8,22 @@
...
+    int res = OpenBCISerialBoard::prepare_session ();
+    if (res != STATUS_OK)
+    {
+        return res;
+    }
+    res = config_board ("d");
+    // cyton sends response back, clean serial buffer
+    unsigned char tmp;
+    while (serial->read_from_serial_port (&tmp, 1) == 1)
+        ;
```
daniellasry, line 24: Would the response be worth logging?

Andrey1994: spdlog can not has configurable EOL symbol behaviour, so we will log each character at new line, or join all chars to temp string and log after that...

PR #42, opened 2020-05-09
improve timestamps for wifi shield

Andrey1994:

Idea - read single package at time, this way we will get more accurate timestamps 

@daniellasry I've tested only ganglion wifi(mostly because I dont want to reconnect wifi shield) would be nice to review carefully cyton and daisy


daniellasry:
I didn't have a chance to test cyton/daisy with my board but the code looks good.

PR #43, opened 2020-05-10
Wifi impedance

Andrey1994:
Should be merged to #42 after that #42 should be merged to master.

It works but very unreliable and I see no ways to make it better. Some commands like 'Z' are not received by device, so there is such code

```
        if (is_cheking_impedance)
        {
            send_config ("Z");
            // call above fails often with message "Lost Connection with OpenBCI Board"
            return STATUS_OK;
        }
```

But based on my experience if this device is sending impedance data any command to device may fail randomly


src/board_controller/openbci/ganglion_wifi.cpp
```diff
@@ -93,6 +242,80 @@ void GanglionWifi::read_thread ()
...
+        int i = 0;
+        // find first non zero byte
+        for (i = 20; i > 2; i--)
+        {
+            if (b[i] != 0)
+            {
+                break;
+            }
+        }
+        std::string asci_value ((const char *)(b + 2), i - 2);
```
daniellasry, line 294: If i remember, the impedance string was terminated with 'Z'. did you have to account for that somewhere?

Andrey1994: For ganglion BLE, yes its Z terminated. For Ganglion WIFI just zeroes

PR #44, opened 2020-05-15
fix ganglion wifi impedance issue

Andrey1994:


src/board_controller/openbci/novaxr.cpp
```diff
@@ -94,6 +94,28 @@ int NovaXR::config_board (char *config)
         safe_logger (spdlog::level::err, "Failed to config a board");
         return BOARD_WRITE_ERROR;
     }
+    if (!is_streaming)
+    {
+        char response = 0;
+        res = socket->recv (&response, 1);
```
daniellasry, line 100: Is there a timeout for this?

Andrey1994: yes, the same as for data packages(5 seconds)

PR #55, opened 2020-05-25
add cyton check

Andrey1994:

@daniellasry could you run it also before merging? I've tested with plain Cyton(have not tested Daisy) and looks like it works. But I am not sure that all boards/firmware versions send the same string if cyton is not connected and not sure about daisy

### DIFF
--- a/src/board_controller/openbci/openbci_serial_board.cpp
+++ b/src/board_controller/openbci/openbci_serial_board.cpp
@@ -29,7 +29,7 @@ int OpenBCISerialBoard::open_port ()
         return PORT_ALREADY_OPEN_ERROR;
     }
 
-    Board::board_logger->info ("openning port {}", serial->get_port_name ());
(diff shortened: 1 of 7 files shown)


src/board_controller/openbci/openbci_serial_board.cpp
```diff
@@ -144,36 +143,68 @@ int OpenBCISerialBoard::prepare_session ()
...
     {
+        delete serial;
+        serial = NULL;
         return port_open;
     }
 
     int set_settings = set_port_settings ();
     if (set_settings != STATUS_OK)
     {
+        delete serial;
```
Andrey1994, line 154: BTW we dont need it here, if session is failed board_controller will call a destructor which will delete serial object(but it may be unclear for somebody)

daniellasry:
I'll test it

daniellasry:
Tested on windows, works with cyton and daisy

PR #61, opened 2020-06-22
Rework FFT calculations, refactor C++ enums and socket helpers

Andrey1994:

Changes:

* add an option to apply window before FFT calculation **Breaking change, new argument in method** 
* make FFT output exactly like in SciPY(positive/negative imag)
* add get_nearest_power_of_two, get_psd, get_log_psd and get_band_power
* add parallel compilation, should be compiled faster now in CI and using bash scripts
* **Convert enums in C++ code to classes to avoid name conflicts(Breaking change for C++)**
* **Fix Streaming Board on multinode**
* Use intermediate buffer for TCP instead SO_RCVLOWAT. recv method may return less bytes than requested and it makes parsing code much more complicated, before to prevent it I used SO_RCVLOWAT but looks like it didnt work well. I even didnt find in docs how it should work together with timeout, to ensure that we dont split package in this PR I added a queue

For PSD it doesnt use moving window like in welch method, but its simple to implement it in user code and I will add it later as well.

Due to some reasons Multicast Socket doesnt work on multinode if you set interface as
```
mreq.imr_interface.s_addr = htonl (INADDR_ANY);
```
Solution which did the trick for me: get local address of PC and use it instead. To get local address I connect to Google DNS


Andrey1994:
int instead unsigned int cause I am not sure that all languages have unsigned int

PR #63, opened 2020-06-28
publish java package to github packages

Andrey1994:

https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-apache-maven-for-use-with-github-packages


Andrey1994:
its very complicated to publish smth to maven central, github package is much easier to configure. But users will need to change settings.xml for maven to work with github packages. Example of settings.xml file is attached just need to change USER and TOKEN by github user and token with an access to github packages

daniellasry:
I'm not too familiar with how java packages are published, but this seems fine

Andrey1994:
in maven central its terrible I spent several days to create issue in their jira, confirm domain name and so on. They force you to add huge amount of plugins and other info to your pom.xml before publishing. And when I finally did it, plugin installation on MacOS in Travis took around an hour and travis killied this job.

So github package looks like the only option unless we find cool devops to configure this pipeline correctly. I dont want to return to it

PR #65, opened 2020-06-29
improve synthetic board

Andrey1994:

* Add one more ppg channel to synthetic board
* Use different amplitudes, noise levels and frequencies for eeg, its useful for testing FFT and related methods. And more like in OpenBCI GUI


Andrey1994:
Agreed, but we should convince others...

daniellasry:
Who is against this?

PR #66, opened 2020-07-05
use dyn lib

Andrey1994:

Idea of this PR - move all weird code and specially code which is outside our control to dyn libraries loaded in runtime manually instead linking. It should be a standard pattern to add third party dependencies which we use as binaries.

As of right now, I wanna fix it for Ganglion(BGLib) and NeuroSDK

It's more code to write and all these function pointers look unclear but there are much more important benefits:

* for BGLib we will be able to reuse BGLib for other BLE devices(if any). If we try to do it right now it will not work because of global functions - callbacks.
* for NeuroSDK it will isolate NeuroSDK issues(specially on MacOS) from other devices. Makes a lot of sense due to upcomming MacOS with ARM processors, wo linking we will not need to wait for NeuroSDK developers to provide MacOS ARM version of their lib
* **Most important agvantage** - wo dependencies as binaries we will be able to compile BrainFlow for any platform wo ifdefs, or custom cmake files


Andrey1994:
Ready for review, I am testing now

Andrey1994:
Update: now it also compiles using Android NDK, but I am having problems with JNA for Android, its more complicated than for PC

PR #79, opened 2020-07-28
add type hints for julia and python

Andrey1994:

* Add type hints for python and type checks for julia
* fix java package on linux

Wanted to add arg validation for matlab but it doesnt work before matlab 2019b

Will be good to double check during review, there is a lot of copypaste

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: bionic
 
 os:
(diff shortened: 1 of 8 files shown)


Andrey1994:
there is a bug in python < 3.5.6 related to type hints, bump min python verion to 3.5.6(it old enough) and move to ubuntu 18 in travis

PR #80, opened 2020-07-31
ML

### LINKED ISSUE #82: Bug in BandPower calculation
BandPower calculation is wrong, it should be an integral instead average of Y values

Andrey1994:
Changes:

* Fix #82 
* drop get_log_psd I realized that it doesnt make any sense
* add get_avg_band_powers to calc band powers across all channels
* add core module for different classifiers
* add regression classifier for Focus
* compile libsvm(not used currently but will be)
* add get_exg_channels


Andrey1994:
still working on docs, examples and minor stuff including small changes in code but I would say its ready for initial review

Andrey1994:
I will move eeg metric and classifier to a class and serialize it like BrainFlowInputParams, it will allow us to add more options later

cpp-package/src/data_filter.cpp
```diff
@@ -214,35 +194,51 @@ std::pair<double *, double *> DataFilter::get_log_psd_welch (
...
         throw BrainFlowException (
-            "nfft is not power of 2", (int)BrainFlowExitCodes::INVALID_ARGUMENTS_ERROR);
+            "Invalid params", (int)BrainFlowExitCodes::INVALID_ARGUMENTS_ERROR);
     }
-    double *ampl = new double[nfft / 2 + 1];
-    double *freq = new double[nfft / 2 + 1];
-    int res = ::get_psd_welch (data, data_len, nfft, overlap, sampling_rate, window, ampl, freq);
+    double *data_1d = new double[cols * channels_len];
+    double *avg_bands = new double[5];
+    double *stddev_bands = new double[5];
```
daniellasry, line 217: no reason for these to be dyamically allocated

Andrey1994: its returned to the user and you can not return a pointer to local variable

daniellasry: ah, that's right

csharp-package/brainflow/brainflow/board_shim.cs
```diff
@@ -236,6 +236,29 @@ public static int[] get_eog_channels (int board_id)
...
+        /// <summary>
+        /// get row indices of EXG channels for this board
+        /// </summary>
+        /// <param name="board_id"></param>
+        /// <returns>array of row nums</returns>
+        /// <exception cref="BrainFlowException">If this board has no such data exit code is UNSUPPORTED_BOARD_ERROR</exception>
+        public static int[] get_exg_channels (int board_id)
+        {
+            int[] len = new int[1];
+            int[] channels = new int[512];
```
daniellasry, line 248: What about these? I don't think they are being returned

Andrey1994: its c#)

daniellasry: Yep Nevermind!
It's been a long day 🙃

Andrey1994:
btw I failed to make valgrind happy about it... but I didnt find any issues

daniellasry:
> btw I failed to make valgrind happy about it... but I didnt find any issues

Is the data being properly released by the caller?

Andrey1994:
as I remember it complained about smth in eeg_metrics.cpp(and code behind it) with uninitialized variables(memory ranges) but I didnt find any unitialized variables

src/data_handler/data_handler.cpp
```diff
@@ -891,18 +882,155 @@ int get_psd_welch (double *data, int data_len, int nfft, int overlap, int sampli
     return (int)BrainFlowExitCodes::STATUS_OK;
 }
 
-int get_log_psd_welch (double *data, int data_len, int nfft, int overlap, int sampling_rate,
-    int window_function, double *output_ampl, double *output_freq)
+int get_avg_band_powers (double *raw_data, int rows, int cols, int sampling_rate, int aply_filters,
+    double *avg_band_powers, double *stddev_band_powers)
```
daniellasry, line 886: Not a big deal, but I would like to see a big function like this broken up into logical chunks that are easier to read and maintain.

For example there's a parallel for loop in here. Could that be it's own function?

src/ml/train/focus_classifiers.py
```diff
@@ -0,0 +1,89 @@
...
+from sklearn.dummy import DummyClassifier
+from sklearn.model_selection import train_test_split, cross_val_score
+
+import brainflow
+from brainflow.board_shim import BoardShim, BrainFlowInputParams, LogLevels, BoardIds
+from brainflow.data_filter import DataFilter, FilterTypes, AggOperations, WindowFunctions, DetrendOperations
+from brainflow.ml_model import BrainFlowMetrics, BrainFlowClassifiers, MLModel
+
+
+def prepare_data ():
```
daniellasry, line 17: I'd say the same about this function. It's too difficult to tell what's going on. 

I see a nested try/except, so i'm sure this could be broken up into logical chunks

PR #92, opened 2020-10-05
Updating Src and High Level Logging

John42506176Linux:
The changes to high level API's and MLmodel will come next just wanted to make sure I'm on track.


src/data_handler/data_handler.cpp
```diff
@@ -18,14 +18,71 @@
 #include "wavelib.h"
 
 #include "FFTReal.h"
-
-
+#include "spdlog/sinks/null_sink.h"
+#include "spdlog/spdlog.h"
+#define LOGGER_NAME "brainflow_logger"
```
Andrey1994, line 23: lets rename it to smth like data_handler_logger or brainflow_data_handler_logger

src/data_handler/inc/data_handler.h
```diff
@@ -49,7 +49,10 @@ extern "C"
 
     SHARED_EXPORT int CALLING_CONVENTION get_avg_band_powers (double *raw_data, int rows, int cols,
         int sampling_rate, int aply_filters, double *avg_band_powers, double *stddev_band_powers);
-
+    // logging methods
+    SHARED_EXPORT int CALLING_CONVENTION set_log_level (int log_level);
+    SHARED_EXPORT int CALLING_CONVENTION set_log_file (char *log_file);
+    SHARED_EXPORT int CALLING_CONVENTION log_message (int log_level, char *message);
```
Andrey1994, line 55: I think this method should be removed and seems like there is no implementation for it in cpp file. Probably you forgot to remove it from header

Andrey1994 (commented):
I like it so far. Also need to add new methods to all bindings. Take a look at BoardShim.py to check how its done for BoardController(and the same for other languages)

Andrey1994:
In MLModel there will be one unclear trick - probably we will want logs in release_session method but since release_session is also called from destructor its not safe to use spdlogger as is. In BoardController there is safe_logger to solve this issue, probably will need to do smth similar

src/data_handler/inc/data_handler.h
John42506176Linux: You were referring only to the log_message method right.

Andrey1994: yes

cpp-package/src/inc/data_filter.h
```diff
@@ -7,11 +7,31 @@
 #include "brainflow_exception.h"
 #include "data_handler.h"
 
+enum class LogLevels : int
```
Andrey1994, line 10: I think its not good to add it here... first of all there is such enum in BoardShim as I remember and more likely it will lead to multiple definitions error. It can be moved from both places to brainflow_constants.h or smth like that or to new header. Or you can include board_shim.h right here if needed

But I think that we can completely remove it from data_filter and ml_model and make set_log_level method private. So, in user API we will keep only enable\disable logger and in this case there is no need for such enum. (ok to hardcode 0, 2 and 6 internally)

cpp-package/src/inc/data_filter.h
```diff
@@ -7,11 +7,31 @@
...
+    LEVEL_ERROR = 4,    /// ERROR
+    LEVEL_CRITICAL = 5, /// CRITICAL
+    LEVEL_OFF = 6       // OFF
+};
 
 /// DataFilter class to perform signal processing
 class DataFilter
 {
 public:
+    static void set_log_level (int log_level);
```
Andrey1994, line 25: I think it can be private and we can keep only enable\disable\enable_dev logs in user API

src/data_handler/data_handler.cpp
```diff
@@ -93,7 +93,7 @@ int perform_lowpass (double *data, int data_len, int sampling_rate, double cutof
     Dsp::Filter *f = NULL;
     if ((order < 1) || (order > MAX_FILTER_ORDER) || (!data))
     {
-        data_logger->error ("Order must be from 1-8 and data cannot be empty. Order:{} , Data:{}",order,data);
+        data_logger->error ("Order must be from 1-8 and data cannot be empty. Order:{} , Data:{}",order,*data);
```
Andrey1994, line 97: lets print boolean value for data, you can write smth like data != NULL

src/data_handler/data_handler.cpp
```diff
@@ -140,7 +140,7 @@ int perform_highpass (double *data, int data_len, int sampling_rate, double cuto
 
     if ((order < 1) || (order > MAX_FILTER_ORDER) || (!data))
     {
-        data_logger->error ("Order must be from 1-8 and data cannot be empty. Order:{} , Data:{}",order,data);
+        data_logger->error ("Order must be from 1-8 and data cannot be empty. Order:{} , Data:{}",order,*data);
```
Andrey1994, line 144: the same here and in methods below  for data

tests/cpp/signal_processing_demo/src/signal_filtering.cpp
```diff
@@ -21,7 +21,7 @@ int main (int argc, char *argv[])
     int board_id = (int)BoardIds::SYNTHETIC_BOARD;
 
     BoardShim::enable_dev_board_logger ();
-    DataFilter::enable_dev_board_logger ();
+    DataFilter::enable_dev_board_logger();
```
Andrey1994, line 24: lets keep code style, you can enable clang-format and it will format all your code automatically on save(https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style) I use sublime with clang-format plugin for example

CMakeLists.txt
```diff
@@ -296,7 +296,9 @@ endif (UNIX AND NOT ANDROID)
 # link android logging library
 if (ANDROID)
     find_library (log-lib log)
+    target_link_libraries (${ML_MODULE_NAME} PRIVATE log)
     target_link_libraries (${BOARD_CONTROLLER_NAME} PRIVATE log)
+    target_link_libraries (${DATA_HANDLER_NAME} PRIVATE ${DSPFILTERS} ${WAVELIB} log)
```
Andrey1994, line 301: no need to link dspfilter and wavelib here, in cmake target_link_libraries works incrementally

John42506176Linux:
Also, so I'm thinking of add a Base classifier.cpp file which will hand all the log methods just so we don't have so much redundancy especially when we add more classifiers later on.

src/data_handler/data_handler.cpp
```diff
@@ -19,25 +19,82 @@
...
+{
+#ifdef __ANDROID__
+    data_logger->error ("For Android set_log_file is unavailable");
+    return (int)BrainFlowExitCodes::GENERAL_ERROR;
+#else
+    try
+    {
+        spdlog::level::level_enum level = data_logger->level ();
+        data_logger = spdlog::create<spdlog::sinks::null_sink_st> (
+            "null_logger"); // to dont set logger to nullptr and avoid raice condition
```
Andrey1994, line 49: typo in race condition

src/ml/concentration_knn_classifier.cpp
```diff
@@ -5,12 +5,15 @@
 #include "brainflow_constants.h"
 #include "concentration_knn_classifier.h"
 #include "focus_dataset.h"
+#include "spdlog/sinks/null_sink.h"
```
Andrey1994, line 8: I think this header is not used here

src/ml/concentration_regression_classifier.cpp
```diff
@@ -5,6 +5,7 @@
 #include "brainflow_constants.h"
 #include "concentration_regression_classifier.h"
 
+#include "spdlog/sinks/null_sink.h"
```
Andrey1994, line 8: and here its not needed

src/ml/inc/base_classifier.h
```diff
@@ -1,20 +1,43 @@
...
 public:
+    static std::shared_ptr<spdlog::logger> ml_logger;
+    static int set_log_level (int log_level);
+    static int set_log_file (char *log_file);
+
     BaseClassifier (struct BrainFlowModelParams model_params) : params (model_params)
     {
     }
 
     virtual ~BaseClassifier ()
```
Andrey1994, line 20: in destructor of base class and in destructors of subclasses skip_logs variable should be set to true. Its important because spdlog doesnt work if its called from destructors at the program end

Andrey1994 (commented):
looks good. The most important missed part is skip_logs set to True in destructors of subclasses before calling release. And high level api changes

src/ml/inc/base_classifier.h
Andrey1994: https://github.com/gabime/spdlog/issues/163#issuecomment-171788629

python-package/brainflow/data_filter.py
```diff
@@ -286,6 +298,65 @@ def __init__ (self):
...
+        """disable BrainFlow Logger"""
+        cls.set_log_level (LogLevels.LEVEL_OFF.value)
+
+    @classmethod
+    def enable_dev_board_logger (cls) -> None:
+        """enable BrainFlow Logger with level TRACE, uses stderr for log messages by default"""
+        cls.set_log_level (LogLevels.LEVEL_TRACE.value)
+
+    @classmethod
+    def log_message (cls, log_level: int, message: str) -> None:
```
Andrey1994, line 329: there is no log_message in DataHandler

python-package/brainflow/data_filter.py
```diff
@@ -286,6 +298,65 @@ def __init__ (self):
 class DataFilter (object):
     """DataFilter class contains methods for signal processig"""
 
+    @classmethod
+    def set_log_level (cls, log_level: int) -> None:
```
Andrey1994, line 302: lets make it private to match other bindings and remove doc strings

src/ml/inc/concentration_knn_classifier.h
```diff
@@ -11,6 +11,10 @@
 class ConcentrationKNNClassifier : public BaseClassifier
 {
 public:
+    static std::shared_ptr<spdlog::logger> ml_logger;
+    static int set_log_level (int log_level);
+    static int set_log_file (char *log_file);
```
Andrey1994, line 16: why is it here? I think it must be only in BaseClassifier

src/ml/inc/concentration_regression_classifier.h
```diff
@@ -6,12 +6,17 @@
 class ConcentrationRegressionClassifier : public BaseClassifier
 {
 public:
+    static std::shared_ptr<spdlog::logger> ml_logger;
+    static int set_log_level (int log_level);
+    static int set_log_file (char *log_file);
```
Andrey1994, line 11: the same here

Andrey1994 (commented):
lets merge https://github.com/John42506176Linux/brainflow/pull/1 to add julia logs and fixes to this PR

Andrey1994 (approved):
looks great! I will merge it after CI

PR #97, opened 2020-10-11
 cannot deference value-initialized deque iterator

nosa999:
in this repository, I changed some codes that related to ganglion board for connecting my BLE board to brainflow and then to OpenBCI-GUI. but I get error when I am running code

image


Andrey1994:
problem solved, for bitalino code lets create a new PR

PR #98, opened 2020-10-11
ganglion improvements

Andrey1994:


Andrey1994:
I thought a little more and seems like ticketlock is not required here, it can be plain mutex or spinlock(but its not an issue either). Idea was that mutex\spinlock doesnt ensure FIFO order(in fact doesnt ensure any order) but here it doesnt matter

daniellasry (approved):

Issue #103, opened 2020-10-12
add OYMotion GForcePro armband to brainflow


Andrey1994:
http://www.oymotion.com/en/product32/149

PR #105, opened 2020-10-13
Oymotion gforce pro armband

Andrey1994:
#103


Andrey1994:
It finally works, I can see emg data.

TODO:

* add package id
* check units which it returns(not sure that its uV)
* there is sampling rate issue, currently is much slower than it should be

PR #107, opened 2020-10-14
rename logging methods and parallel band power calculation

Andrey1994:

### DIFF
--- a/src/data_handler/data_handler.cpp
+++ b/src/data_handler/data_handler.cpp
@@ -5,6 +5,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <string>
+#include <thread>
 #include <vector>
 
 #include "brainflow_constants.h"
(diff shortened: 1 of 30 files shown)


Andrey1994:
would be nice to review `get_avg_band_powers` method carefully

Andrey1994:
with openmp instead c++ threads it looked much better but I didnt find a way to link libomp statically

src/data_handler/data_handler.cpp
```diff
@@ -1129,4 +1115,68 @@ int get_avg_band_powers (double *raw_data, int rows, int cols, int sampling_rate
...
+{
+    for (int i = row_num_start; i < row_num_stop; i++)
+    {
+        double *ampls = new double[nfft / 2 + 1];
+        double *freqs = new double[nfft / 2 + 1];
+        double *thread_data = new double[cols];
+        memcpy (thread_data, raw_data + i * cols, sizeof (double) * cols);
+
+        if (aply_filters)
+        {
```
John42506176Linux, line 1135: Might want to add a log method to state which method failed for the exit code as I don't think the log methods in those methods specify their methods. Not sure if it would be overly clunky though, so it's your call.

John42506176Linux (approved):
Looks fine to me, left a comment about one thing but that's about it,

PR #129, opened 2020-11-10
Unrecognised Escape Sequence compiler Warnings

stellarpower:
I get warnings from g++ about unrecognised escape sequence (`\.`) - I'm presuming the regular expression has been copied but without doubling the backslashes to get a literal backslash in the C-style string literal. 

For confirmation, I haven't tested this and I don't have any hardware I can use to test. I'm just going on the static code and what the build process has told me.

### DIFF
--- a/src/board_controller/openbci/openbci_wifi_shield_board.cpp
+++ b/src/board_controller/openbci/openbci_wifi_shield_board.cpp
@@ -362,7 +362,7 @@ std::string OpenBCIWifiShieldBoard::find_wifi_shield ()
             {
                 std::string response ((const char *)b);
                 safe_logger (spdlog::level::trace, "Recived package {}", b);
-                std::regex rgx ("LOCATION: http://([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)");
+                std::regex rgx ("LOCATION: http://([0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]+)");
                 std::smatch matches;
                 if (std::regex_search (response, matches, rgx) == true)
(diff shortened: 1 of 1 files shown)


Andrey1994 (changes requested):
Hello and thanks!

But this regexp is used to find ip address, so format for it is  like 192.168.4.1 with double slash you probably trying to find smth like 192\\.168\\.4\\.1 and its wrong.

I suppose correct fix is to remove slash

I think I added \ before . because I supposed that . is a special character in regexp but seems like it is not

stellarpower:
I think the regular expression is right, for something like grep, because `[0-9]+.` matches against `124P`, and `[0-9]+\.` correctly removes the specisal meaning of `.` and makes it a literal full stop. 

However, the backslash as part of a string literal in C++ is used to introduce an escape character like in `\n`, so the compiler is warning it doesn't understand the special character `\.` - I would have to check documentation, but I imagine this is probably in the realm of undefined behaviour, and the generated code may simply ignore the backslash or ignore the character altogether. In regular expressions in C/C++, you need to escape the first backslash to get a literal backslash in the string; this is then passed to the regular expression library as a backslash that applies to the next character, i.e. the `.`. Thus, to get a literal backslash inside a regular expression in C++ would need `\\\\`

See for example [this odcument in Qt](https://doc.qt.io/qt-5/qregularexpression.html#introduction), or the example in the docs for [std::regex](https://en.cppreference.com/w/cpp/regex#Example)

Andrey1994 (approved):
seems like you are right, I checked and it works

Issue #151, opened 2020-12-02
Refactor Julia Code


matthijscox:
Good first issue for me; to refactor the Julia code with my current Julia knowledge.

I've started reading the Julia code, the first line I encounter is 
```
AnyIntType = Union{Int8, Int32, Int64, Int128, Int}
```
This is just weird... `Int` is the abstract supertype of the other types, you shouldn't have to declare a `Union`. Just try it in a Julia REPL:
```
julia> Int64 <: Int
true
```
So we can replace `get_timestamp_channel(board_id::AnyIntType)` simply with `get_timestamp_channel(board_id::Int)`

I can probably find more improvements, maybe even performance enhancements.

Andrey1994:
Nice, I didnt know it

Andrey1994:
Also, about refactoring, I dont like this https://github.com/brainflow-dev/brainflow/blob/master/julia-package/brainflow/src/board_shim.jl#L70 but if I move it to a function julia crashes

And enums in Julia as far as I understand work like defines in C++ and I dont like it at all and can not be duplicated in entire module. E.g https://github.com/brainflow-dev/brainflow/blob/master/julia-package/brainflow/src/data_filter.jl#L35 

Maybe there is better alternative to built-in enums

Andrey1994:
And there is a task to add Julia tests to CI system, its a part of https://github.com/brainflow-dev/brainflow/issues/146

matthijscox:
Yeah, I was looking at the `ccall` you were doing. I believe it's solved if you use a `const LIBRARY_PATH = "./lib/DataHandler.so"`
At least that's how we did it in another project.

Did you try using `const` ? If not, I will include this in my refactoring.

I also disliked enums, they are hardly used in Julia, nstead of enums everyone always just use subtypes:
```
abstract type DetrendOperations end
struct ConstantDetrend <: DetrendOperations end
struct LinearDetrend <: DetrendOperations end
```
This leads to maximum use of Multiple Dispatch, you can just dispatch on the desired type whenever you write a detrend function. I'll include it in the refactoring.

Andrey1994:
"Did you try using const ? If not, I will include this in my refactoring."
No, but if its a function which returns library name based on current OS can it be const?

My idea was smth like
```
lib_name = get_lib_name()
```

matthijscox:
yes, you can initialize the `const` that way. 

This was my first idea:
```
[...]

PR #154, opened 2020-12-03
[WIP] 151 julia refactor

matthijscox:
Related to issue [151](https://github.com/brainflow-dev/brainflow/issues/151)

Refactoring so far:
* replaced library paths with `const` variables. 
* replaced `AnyIntType` Union with `Integer`
* replaced `AnyDoubleType` Union with `Float64`
* started removing `Integer()` calls in API functions

Primary thing to test/discuss is the library path replacement. At least it would be good if Andrey checks it doesn't cause the issue he had before. But right now I hardcoded the path to be `julia-package/brainflow/lib`, so you'll have to place the library files there.

Perhaps we should first do the automated artifact deployment before merging this refactoring, then we can use `const INTERFACE_PATH = artifact"brainflow"` which will automatically install the libraries in the julia artifact directory and provide the path.


Andrey1994:
I like changes so far

Andrey1994:
Ignore failed pipelines, its an error in CI, should not try to push artifacts to AWS for pull requests

Andrey1994:
I've fixed CI issue for PRs and added automated tests for Julia to CI pipelines here https://github.com/brainflow-dev/brainflow/pull/155

If you port this changes to your branch you will see results of tests running in CI runner for all OSes(you can just merge master branch to your branch)

matthijscox:
I've done some heavy refactoring, greatly simplifying the code in my opinion. Especially added some Julia code generation patterns to remove a bunch of redundancy.

Most Enums are replaced by Julia types, see if you prefer that. It gives you more configuration at least, like custom JSON conversion or Integer conversion.

CI failed because the build server uses Julia 1.2 and I set compat to >=1.4. I recall I needed at least 1.3 for something, but I'll try the CI with 1.2 compat. I think it's the Pkg.Artifact that requires 1.3, which I am not using yet.

julia-package/brainflow/test/julia_tests.jl
```diff
@@ -0,0 +1,48 @@
+using Test
+using brainflow
+using JSON
+
+@testset "prepare_session error" begin
+    # use a board that is not connected
+    params = brainflow.BrainFlowInputParams()
+    board_shim = brainflow.BoardShim(brainflow.CYTON_BOARD, params)
```
Andrey1994, line 8: I think it should be replaced by Synthetic Board, Cyton will not work wo real hardware. Also prepare without release will lead to memory leak

julia-package/brainflow/test/transforms.jl
```diff
@@ -1,12 +1,12 @@
-import brainflow
+using brainflow
```
Andrey1994, line 1: Just curios whats the difference between using and import

Andrey1994:
I will patch CmakeLists file to copy compiles libs to a folder julia-package/brainflow/lib it will fix broken CI

Andrey1994:
We can change Julia version in CI, I dont want to set the newest one cause it will limit amount of users but we definitely can move from 1.2 to 1.3

julia-package/brainflow/test/julia_tests.jl
Andrey1994: Sorry I misunderstood it, its a test which should fail right? Then its ok

matthijscox:
> I will patch CmakeLists file to copy compiles libs to a folder julia-package/brainflow/lib it will fix broken CI

I hope it's not too much effort for you.
If you install Julia 1.3 and make a tarball I can instead make the artifacts work.

julia-package/brainflow/test/julia_tests.jl
matthijscox: Correct

julia-package/brainflow/test/transforms.jl
matthijscox: https://docs.julialang.org/en/v1/manual/modules/#Summary-of-module-usage

`import` only brings the package into the scope. `using` also brings the explicitly exported functions and types into scope.

Andrey1994:
I already did it. Take a look at https://github.com/brainflow-dev/brainflow/pull/159/files

Andrey1994:
I like changes, lets fix CI pipeline(error there is not related to libs) and merge it

Andrey1994:
https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/run_windows.yml#L220
https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/run_unix.yml#L98

feel free to change these commands in CI if needed

julia-package/brainflow/src/board_shim.jl
```diff
@@ -38,620 +36,172 @@ end
...
+    function BoardShim(id::Integer, params::BrainFlowInputParams)
         master_id = id
         if id == Integer(STREAMING_BOARD) || id == Integer(PLAYBACK_FILE_BOARD)
             try
                 master_id = parse(Int, id)
             catch
-                throw(BrainFlowError("you need to provide master board id to other_info field"))
+                throw(BrainFlowError("you need to provide master board id to other_info field", 1))
             end
         end
```
matthijscox, line 136: I dont understand the purpose of this part. And I think it would even fail:
```
julia> parse(Int, Int32(5))
ERROR: MethodError: no method matching parse(::Type{Int64}, ::Int32)
````

Comment (Andrey1994) [reply julia-package/brainflow/src/board_shim.jl:136]: BrainFlowError Defined like
```
struct BrainFlowError <: Exception
    msg::String
    ec::Int32
end
```
First arg is a string and second one is one of exit codes, I think instead 1 need to use INVALID_ARGUMENTS_ERROR from https://github.com/brainflow-dev/brainflow/blob/master/julia-package/brainflow/src/exit_codes.jl#L16

Idea is to throw an exception of known type(BrainFlowError)

Comment (Andrey1994) [reply julia-package/brainflow/src/board_shim.jl:136]: it parses other_info field of BrainFlowInputParams and its a string, for some boards you need to provide int number to this string field.

Comment (matthijscox) [reply julia-package/brainflow/src/board_shim.jl:136]: Yes I get the exception. I mean the purpose of the `if` statement at all.

Input `id` is an `Integer`, so no need to try to parse it again into an `Int` right?
```
function BoardShim(id::Integer, params::BrainFlowInputParams)
        master_id = id
        if id == Integer(STREAMING_BOARD) || id == Integer(PLAYBACK_FILE_BOARD)
            try
                master_id = parse(Int, id)
            catch
                throw(BrainFlowError("you need to provide master board id to other_info field", 1))
            end
        end
        new(master_id, id, JSON.json(params))
    end
```

Can we just remove the `if` and copy the `id` into the `master_id`:
```
function BoardShim(id::Integer, params::BrainFlowInputParams)
    new(id, id, JSON.json(params))
end
```
Or maybe add another constructor for the String. This should already throw a good exception if the string cannot be converted to an int.
```
function BoardShim(master_id::AbstractString, id::Integer, params::BrainFlowInputParams)
    int_master_id = parse(Int, master_id)
    new(int_master_id , id, JSON.json(params))
end
```

**Comment by matthijscox** [comment]
CI working 🎈

File: julia-package/brainflow/src/board_shim.jl (hunk 3 above)
Comment (matthijscox) [reply julia-package/brainflow/src/board_shim.jl:136]: Are there examples from these special boards, and how they should use the input_json generated from the BrainFlowInputParams?

Comment (Andrey1994) [reply julia-package/brainflow/src/board_shim.jl:136]: Oh, seems like I made a mistake. It should be like
```
 if id == Integer(STREAMING_BOARD) || id == Integer(PLAYBACK_FILE_BOARD)
            try
                master_id = parse(Int, BrainFlowInputParams.other_info)
            catch
                throw(BrainFlowError("you need to provide master board id to other_info field", 1))
            end
        end
```

matthijscox: fixed

Andrey1994:
Great!
Lets add Julia packaging to another PR. Are there any other changes you want to add here? I think it's ok to merge it

matthijscox:
I am happy with the changes, it's a big chunk of refactoring. Let's merge.

In the another PR, we should also think about the Julia package registration and Julia package version (should it be the same or different from the overall brainflow version?).

Andrey1994:
Definitely the same. I will take a final look at this change and merge
Now publishing is done here https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/deploy.yml#L149

For some languages I do it manually but prepare package for publishing in CI

Andrey1994 (approved):

Issue #144, opened 2020-11-25
Build java libraries for jdk 8 and 11


Andrey1994:
In java9 modules were added, as of right now brainflow uses java 8 and has no modules. It makes it hard to use it in java 11 apps . At the same time we need to keep java 8 version for existing apps which do not support newer java versions.

We can add modules-info.java file for java > 8 and build 2 versions of java package in CI

PR #157, opened 2020-12-06
Add Java 11 and Java 8 Jars

Andrey1994:
https://github.com/brainflow-dev/brainflow/issues/144


Andrey1994:
Java 8 jars will not be available for public, in github packages we will use Java 11 libs and we will publish this jar to release page.

Also, from now on java modules will be used

PR #168, opened 2020-12-16
Allow using BoardIds as BoardShim argument

imachug:
BoardIds used to be a enum.Enum, not enum.IntEnum, which caused problems when passing the ID to native code. Making it IntEnum instead makes the following code work, which is the 'right' code as hinted by documentation:

```python
board = BoardShim(BoardIds.SYNTHETIC_BOARD, params)
```

### DIFF
--- a/python-package/brainflow/board_shim.py
+++ b/python-package/brainflow/board_shim.py
@@ -15,7 +15,7 @@ from nptyping import NDArray, Float64
 from brainflow.exit_codes import BrainflowExitCodes
 
 
-class BoardIds (enum.Enum):
+class BoardIds (enum.IntEnum):
     """Enum to store all supported Board Ids"""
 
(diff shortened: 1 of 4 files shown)


Andrey1994:
Hi,

I think enums should be consistent everywhere, so if we want to allow board ids without .value all other enums in python binding should be updated the same way, and even more ideally it should be the same in other bindings

imachug:
That is a good idea, thanks for the recommendation. I've replaced all instances of enum.Enum with enum.IntEnum in Python code. Sorry for multiple commits but I'm having trouble using CLI git at the moment, feel free to squash them. I have also not updated other bindings because unfortunately I don't know other languages well enough to judge if any changes have to be done.

Andrey1994 (approved):

Andrey1994:
Looks good, thanks!

Issue #167, opened 2020-12-15
Add CI pipeline which will test releases


Andrey1994:
We need to confirm that released package works everywhere, so we need to run it for all possible OSes and different python\java\etc versions, inside and outside docker container and so on.  It can be a copypaste from run_unix and run_windows workflows but with more OSes, and with packages downloaded from nuget\pypi\github release page\etc instead compiled directly on machine

Andrey1994:
And we need to run it with and without openmp installed

PR #172, opened 2020-12-18
Artifacts support

matthijscox:
Added artifact"brainflow" support
And a try/catch to support the unpack error on Windows for now

TODO:

- Project.toml update 
  - update version nr to be equal to overall brainflow version
  - update compat to Julia = "1.3" (needed for Artifacts)
- Automatically update Project.toml version in CI
- Automatically update Artifact.toml in CI 
  - update `src/brainflow_url.jl` and 
  - call `generate_brainflow_artifact.jl` script
- Register with JuliaRegistrator. 
  - comment on a commit `@JuliaRegistrator register subdir=julia-package/brainflow`
  - Can be made automatic in CI?
- update Julia brainflow installation documentation
- other things?


matthijscox:
Ok, artifact installation is working.
So `using brainflow` now automatically downloads and installs the c libraries if they are not yet installed.

Libraries are installed in a custom julia folder
You can see the artifact folder (assuming /julia-package/brainflow is your working dir):
```
using Pkg.Artifacts
artifact_path = artifact"brainflow"
```

julia-package/brainflow/Artifacts.toml
```diff
@@ -0,0 +1,7 @@
+[brainflow]
+git-tree-sha1 = "1681bc5c836c0465e02c42dce71db9572d723b6e"
+lazy = true
+
+    [[brainflow.download]]
+    sha256 = "92bd7485de8ccbd19d27a9b203e8176806269415b1087c767bb36539efe063d9"
+    url = "https://github.com/brainflow-dev/brainflow/releases/download/3.7.2/compiled_libs.tar"
```
Andrey1994, line 7: I will think how to update it, maybe need to check how its used in Registrator and what exactly it does

julia-package/brainflow/generate_brainflow_artifact.jl
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    return brainflow_hash
+end
+
+cd(@__DIR__)
+include("src/brainflow_url.jl")
+add_brainflow_artifact!(brainflow_url())
+
+# # try it out; remove artifact and try to retrieve it again
```
Andrey1994, line 45: lets remove these comments

julia-package/brainflow/src/c_interfaces.jl
```diff
@@ -1,3 +1,39 @@
...
+
+    rm(download_path)
+
+    return brainflow_hash
+end
+
+function get_brainflow_artifact_path()
+    # we have an issue with unpack() on Windows, so wrote a custom download
+    # https://discourse.julialang.org/t/unable-to-automatically-install-artifact/51984/2
+    artifacts_toml = find_artifacts_toml(@__DIR__) # is there a better way? this makes brainflow non-relocatable I believe.
```
Andrey1994, line 26: it seems like it always gets libs from the release page instead local folder. I think we can add a check like 

```
if (linux)
    file_name = './lib/libBoardController.so'
if (win)
    file_name = '.\\lib\\BoardController.dll'
if (mac)
    file_name = './lib/libBoardController.dylib'

if (!exists(file_name))
    # download artifacts
```

Will it work?

julia-package/brainflow/Artifacts.toml
matthijscox: Registrator doesnt use this file. Artifacts.toml is only used for installing and finding the binary files. 

Actually the url is now meaningless as I worked around the unpack() issue by making a manual download, but I do want to find out why we have the unpack issue and use the native artifact download functionality. We can accept the tech debt and fix it in another PR though, if you are OK?

julia-package/brainflow/src/c_interfaces.jl
matthijscox: No actually, if the artifact exists in the local folder it is not downloaded:

```
    if artifact_exists(brainflow_hash)
        return artifact_path(brainflow_hash)
```

Note however that we now move away from the `./lib/` folder into a Julia maintained artifact folder

julia-package/brainflow/generate_brainflow_artifact.jl
matthijscox: done

julia-package/brainflow/Artifacts.toml
Andrey1994: its ok, I was not going to change it in this pr. If its not used we probably can keep as is and dont fix it later

julia-package/brainflow/src/c_interfaces.jl
Andrey1994: "Note however that we now move away from the ./lib/ folder into a Julia maintained artifact folder"
Yeah I realized it, can we mix these 2 options? I mean if there are local libs load them if there are not load artifacts

Andrey1994: or maybe there is a better way to achieve it

matthijscox: hmm, yeah we could do this. But what to use in the CI then? Use the develop ./lib path, or the automatically installed libraries from the release url? I prefer to test what the user does (so use the released libraries), but then CI fails while developing until you release the libraries officially. Because if you use ./lib in CI, then after releasing you'll have to manually update CI to check if the automatic download process works on all platforms.

This is why I still prefer a more decoupled deployment process. First work on the c++, when that is fine and tested you (internally) deploy the libraries, then use the new url in the Julia/MATLAB/etc bindings, then we update those bindings, then release those bindings. 

I guess you do internal development only on Python right now anyway? Where you quickly compile c++ libraries, then check how they feel in Python, then update c++ again to improve, repeat.

Andrey1994: "But what to use in the CI then? Use the develop ./lib path, or the automatically installed libraries from the release url?"
It should be two different cases and two different CI runs, in existing workflows we need to test local libs(and there we already precompile them), because existing workflow mostly for developers. I have a task to test released packages(https://github.com/brainflow-dev/brainflow/issues/167) and for julia it will test package from Artifacts

Andrey1994: "I guess you do internal development only on Python right now anyway? Where you quickly compile c++ libraries, then check how they feel in Python, then update c++ again to improve, repeat."

No, its mostly in Cpp but for new functions need to update all bindings at once, so it includes Julia too. Its way more better to do it in single git commit\PR

Andrey1994: Also, another reason to add the ability to use local libs - performace. Precompiled libraries dont have openmp(because users will need to install system packages and other dependencies and some of them dont need it). So, existing policy is that precompiled libs have no OpenMP parallelization and if you want to increase perf - compile from the source with openmp enabled.

In a long term this policy more likely will be changed and I will split it by two packages - with and without openmp

matthijscox: very well, if we get 2 CIs we can split.
I'll add the ./lib back in then.

matthijscox:
I added back `./lib` usage for developers. It takes precedence over the artifact.

Assuming CI passes, shall we first merge and register brainflow to see if the registrator process works? We can then consider how to fully automate the CI for new releases. 

So we would move these items to a next issue:
- Automatically update Project.toml version in deployment
- Automatically update Artifact.toml in deployment
  - update src/brainflow_url.jl and
  - call generate_brainflow_artifact.jl script
- Automate JuliaRegistrator process in deployment

Andrey1994:
yes, I think we can merge it and add everything else to next PRs

Issue #169, opened 2020-12-17
Create dockerfile with all packages preinstalled


Andrey1994:
Docker Image Specs:

* Use ubuntu 20 as a base image
* Install libgomp
* Make sure there is python3 and jdk >= 11
* Install julia and R
* Compile BrainFlow with -DUSE_OPENMP=ON(check #163 )
* Install all bindings from the local folder
* Also lets add some useful packages which are not required by BrainFlow(mne-python, pandas, etc)

For compilation add an option to provide git tag\commit\branch to checkout

I will update CI pipelines to build and publish image automatically later, for now just need to create a DockerFile

Stig124:
I'm actually working on it [https://github.com/Stig124/brainflow/tree/dockerfile](here)

Stig124:
I have managed to do something, I can't find a way to build R/Julia/Matlab binding procedurally, the other binding works great

Andrey1994:
Great!

Dont need to add matlab and I think its impossible at least because its not public software.

To prepare R environment you need to install libxml2-dev and r-base from package manager as I remember(maybe a few more), and before R package installation need to install python package. To install R dependencies you can use smth like

```
sudo -H R --vanilla -e 'install.packages("knitr", repos="http://cran.us.r-project.org")'
sudo -H R --vanilla -e 'install.packages("reticulate", repos="http://cran.us.r-project.org")'
```

For Julia you can take a look at this https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/run_unix.yml#L98 it access Julia terminal from bash

Feel free to create PR and I will take a look

PR #174, opened 2020-12-18
Dockerfile for Brainflow

Stig124:
In response to #169, here is a *(beta)* Dockerfile with everything (except R, Julia and Matlab binding) preinstalled  

Possible fixes : 
- Adding non-root user and install under his name if needed
- `--single-branch` may be unnecessary
- Adding non-essential packages

[Apt-Fast](https://github.com/ilikenwf/apt-fast) is used because it speeds up so dramatically the package install process
`--build-arg` build arguments are :
- branch
- tag
(usage `docker build --build-arg xxxx=xxxx .)


tools/build_docker.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
Andrey1994, line 1: Seems like this file is not needed, its the same as for linux wo docker

Stig124: Actually it's not, it enables the `-DUSE_OPENMP=ON` flag in cmake

Andrey1994: Now there are bash files with openmp https://github.com/brainflow-dev/brainflow/blob/master/tools/build_linux_omp.sh and no need to rebuild this code in docker since it should be precompiled there

Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+ARG tag
+ARG branch
+
+RUN if [ -z "${tag}" ] || [ -z "${branch}" ]; then git clone "https://github.com/brainflow-dev/brainflow.git"; elif [ ! -z "${tag}" ]; then git clone --branch "${tag}" --single-branch "https://github.com/brainflow-dev/brainflow.git"; else git clone --branch "${branch}" --single-branch "https://github.com/brainflow-dev/brainflow.git"; fi
+
+# Python Binding
+WORKDIR /root/brainflow
+RUN python3 -m pip install --user virtualenv
+RUN python3 -m virtualenv venv
+RUN bash ./tools/build_linux.sh
```
Andrey1994, line 45: I think it should run the version with openmp

Andrey1994:
It looks good, I will test it during the weekend

Andrey1994:
Not related to this PR but did you test it with real devices? I suppose for devices which use tcp\ip need to run docker with `--network host`, right? Serial ports should work from the box I guess

Stig124:
> 
> 
> Not related to this PR but did you test it with real devices? I suppose for devices which use tcp\ip need to run docker with `--network host`, right? Serial ports should work from the box I guess

I absolutely didn't tested anything here, just did take care that the build process go through without error

Stig124:
@Andrey1994 Thanks for the help, I was able to install R and Julia binding, I'll tidy the Docker file and add a commit

Stig124:
There you go @Andrey1994, build normally (also removed --single-branch) with every binding except matlab
If needed I can make a non-root version of that

Andrey1994:
Thanks!

PR #176, opened 2020-12-19
Fix build on Linux, .gitignore, warnings

### LINKED ISSUE #119: Fix compilation warnings

imachug:
This fixes most warnings (except third-party packages and unused values in stubs) and so probably closes #119. FWIW, the warnings were from GCC, not MSVC which I don't have atm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -353,16 +353,22 @@ docs/Doxyfile*
 java-package/brainflow/target
 java-package/brainflow/.classpath
 java-package/brainflow/.project
-csharp-package/brainflow/brainflow/lib/brainflow_boards.json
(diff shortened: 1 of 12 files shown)


Andrey1994:
Looks good and thanks for starting it. Idea of #119 is to fix all of them for all compilers and enable warnings as errors check in CI, but it can be iterative process, no need to fix all of them in single PR.

Take a look at this and configure clang-format on your system to fix code style a little https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style 

There was a bug in clang-format CI check, its fixed now, if you merge master branch to your branch you will see that clang-format check fails and you will also see git diff

PR #190, opened 2020-12-26
Added ARM little endian to oscpp library

p-riebs:
This will allow compiling and building oscpp library in ARM based processors. In a majority of cases ARM is little endian.

OSCPP wasn't able to find the correct endian type on a alpine docker image on a raspberry pi 4 and I was getting this error:
`The file oscpp/endian.hpp needs to be set up for your CPU type docker`

I see that there was recent work on the Dockerfile and if you want to I can clean up my alpine Dockerfile and submit that as well. See: https://github.com/p-riebs/brainflow/blob/alpine_dockerfile/Docker/alpine/Dockerfile

### DIFF
--- a/third_party/oscpp/include/oscpp/detail/endian.hpp
+++ b/third_party/oscpp/include/oscpp/detail/endian.hpp
@@ -71,7 +71,7 @@
     defined(__ia64__) || defined(_M_IX86) || defined(_M_IA64) ||     \
     defined(_M_ALPHA) || defined(__amd64) || defined(__amd64__) ||   \
     defined(_M_AMD64) || defined(__x86_64) || defined(__x86_64__) || \
-    defined(_M_X64) || defined(__bfin__)
+    defined(_M_X64) || defined(__bfin__) || defined(__arm__)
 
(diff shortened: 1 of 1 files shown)


Andrey1994:
Looks good.

I dont think that little endian itself is a problem, little endian used on almost all architectures and we tested brainflow on ARM before. I think its compilation/OS speciffic. Or maybe because of musl instead glibc in Alpine Linux.

"I see that there was recent work on the Dockerfile and if you want to I can clean up my alpine Dockerfile and submit that as well. See: https://github.com/p-riebs/brainflow/blob/alpine_dockerfile/Docker/alpine/Dockerfile"

I think we it will be good to add it to the repo but more likely will not publish it to the Dockerhub, feel free to create another PR for it. Make sure that both Dockerfiles have approximately the same preinstalled packages and the same build arguments and behaviour.

Also, you are welcome to create a test for compilation using musl. I suppose the easiest option for it is to compile inside Alpine docker container. Smth like this https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/run_unix.yml#L55

PR #193, opened 2020-12-27
Fix FFT on power-of-two data

imachug:
nfft = 2^n requires at least 2^n + 1 pieces of data, while 2^n would suffice


Andrey1994:
Could you fix this test https://github.com/brainflow-dev/brainflow/blob/master/tests/cpp/signal_processing_demo/src/band_power.cpp to check it? I mean change `get_board_data` by `get_current_board_data` and request 2^n datapoints.

PR #203, opened 2021-01-04
Data viz post

matthijscox:
@Andrey1994 I finished the post, could you review? I haven't had time to try out the docker building, but I am finished for the evening. I'll see if I can find some time to do that, but if you approve and merge I will also be happy.

Also are you really not taking your mono-repo too far, I needed to switch branches all the time to use brainflow while writing the post ;)


Andrey1994:
I will try it locally and maybe patch a little.

Its quite common to use `gh-pages` branch for website and I think you dont need to switch branches if you are using julia package from the registrator. And average users dont deal with wevsite stuff, so it's ok)

PR #205, opened 2021-01-07
Julia artifact fix

matthijscox:
* Artifacts.toml update
* Small documentation update

### DIFF
--- a/docs/BuildBrainFlow.rst
+++ b/docs/BuildBrainFlow.rst
@@ -103,9 +103,9 @@ BrainFlow is a registered package in the Julia general registry, so it can be in
         import Pkg
         Pkg.add("BrainFlow")
         
-When using BrainFlow for the first time in Julia, the BrainFlow artifact containing the compiled BrainFlow libraries will be downloaded.
+When using BrainFlow for the first time in Julia, the BrainFlow artifact containing the compiled BrainFlow libraries will be downloaded from release page automatically.
 
-BrainFlow developers or expert users may also place compiled libraries into /lib inside the julia package root folder. These will take precedence over the artifact.
+If you compile BrainFlow from source local libraries will take precedence over the artifact.
 
 Docker Image
 --------------
--- a/docs/BuildBrainFlow.rst
+++ b/docs/BuildBrainFlow.rst
@@ -101,9 +101,11 @@ BrainFlow is a registered package in the Julia general registry, so it can be in
     Example: ::
 
         import Pkg
-		Pkg.add("BrainFlow")
+        Pkg.add("BrainFlow")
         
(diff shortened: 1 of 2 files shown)


docs/BuildBrainFlow.rst
```diff
@@ -101,10 +101,12 @@ BrainFlow is a registered package in the Julia general registry, so it can be in
     Example: ::
 
         import Pkg
-		Pkg.add("BrainFlow")
+        Pkg.add("BrainFlow")
         
 When using BrainFlow for the first time in Julia, the BrainFlow artifact containing the compiled BrainFlow libraries will be downloaded.
 
+BrainFlow developers or expert users may also place compiled libraries into /lib inside the julia package root folder. These will take precedence over the artifact.
```
Andrey1994, line 108: lets change this sentence.

When you compile brainflow from the source cmake automatically places all libs there you dont need to copypaste libs

julia-package/brainflow/Artifacts.toml
```diff
@@ -1,7 +1,7 @@
 [brainflow]
-git-tree-sha1 = "1681bc5c836c0465e02c42dce71db9572d723b6e"
+git-tree-sha1 = "551bbeee43bcec215c6dab34727a9f87bac06b09"
```
Andrey1994, line 2: will it be the same after merge?

docs/BuildBrainFlow.rst
matthijscox: Sure it's in the CI, but I am also using this for local development. And you could switch manually to the OpenMP variant right?

julia-package/brainflow/Artifacts.toml
matthijscox: Why do you expect it to change? It only depends on the 3.8.1 libs.

matthijscox: If we go for a 3.8.2 release then we need to update it again

docs/BuildBrainFlow.rst
Andrey1994: if compiled locally with openmp enabled,  will be improved in next version

julia-package/brainflow/Artifacts.toml
Andrey1994: I thought its git commit, if its sha of compiled libs then what is sha256 = "520b54a5d8bd9247bd975ccbb73347f46129592acc6b80853a144d95a223ec4a" a few lines below it? They are not the same

Andrey1994:
I've fixed artifacts during adding julia to CI and seems like now it works(https://github.com/brainflow-dev/brainflow/actions/runs/470278824)

I dont understand how it works... I mean in fact we changed smth in already released package without reregistering it. In all languages I know its not allowed

Issue #110, opened 2020-10-18
Implement Common Spatial Patterns algorithm


Andrey1994:
https://en.wikipedia.org/wiki/Common_spatial_pattern

Andrey1994:
Python reference impl:
```
def train_CSP(epo, mrk_class):
"""
Usage:W, d = trainCSP(epo, mrk_class)Parameters:epo: a 3D array of segmented signals (samples x channels x epochs)mrk_class: a 1D array that assigns markers to classes (0, 1)Returns:W: matrix of spatial filtersd: vector of generalized Eigenvalues
"""
C=epo.shape[1]
X1=np.reshape(np.transpose(epo[:,:,mrk_class==0], (1,0,2)), (C,-1))
S1=np.cov(X1)
X2=np.reshape(np.transpose(epo[:,:,mrk_class==1], (1,0,2)), (C,-1))
S2=np.cov(X2)
d,W=scipy.linalg.eigh(a=S1, b=S1+S2)
return W,d
```

W is a matrix of filters and d are the corresponding eigenvalues. In this version of CSP, an EV of 0.5 means explained variances equal in both classes (no discriminability) and 1.0 100% for class mrk_class==0 and 0.0 for the other class.

You would then have to select relevant csps, so e.g. the 2-3 best for each class (highest/lowest EVs) or the 3-6 largest EVs in deviation from 0.5.

Tutorial:

https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwj91IGxhubsAhXMGewKHflIB6AQFjAAegQIARAC&url=http%3A%2F%2Fdoc.ml.tu-berlin.de%2Fbbci%2Fpublications%2FBlaTomLemKawMue08.pdf&usg=AOvVaw3rSEI1KATdOZl8gjGLlI2y

Div12345:
Is this to be implemented in Python?

Andrey1994:
No, python version here just for reference. Should be in Cpp like everything else

Div12345:
Could you give me some pointers as to how to start going about this? Like where this would come under, etc.

PR #213, opened 2021-01-24
Add common spatial patterns 

dvaposto:
#110 

Here is the implementation of the CSP filtering. It computes the matrix W (so called filters) in the matrix equation X' = W * X.


.github/workflows/run_windows.yml
```diff
@@ -197,6 +200,9 @@ jobs:
     - name: Downsampling Cpp Test
       run: .\tests\cpp\signal_processing_demo\build\Release\downsampling.exe
       shell: cmd
+    - name: CSP Cpp Test
+      run: .\tests\cpp\signal_processing_demo\build\Release\csp.exe
+      shell: cmd
```
Andrey1994, line 205: lets add test for c# to pipeline

cpp-package/src/inc/data_filter.h
```diff
@@ -58,6 +58,9 @@ class DataFilter
     /// perform wavelet denoising
     static void perform_wavelet_denoising (
         double *data, int data_len, char *wavelet, int decomposition_level);
+    /// get common spatial patterns
```
Andrey1994, line 61: lets add info about args here

java-package/brainflow/src/main/java/brainflow/examples/CSP.java
```diff
@@ -0,0 +1,34 @@
...
+    public static void main (String[] args) throws Exception
+    {
+        int n_channels = 2;
+        double[] labels = {0, 1};
+        double[][][] data = {
+                            {{6, 3, 1, 5}, {3, 0, 5, 1}},
+                            {{1, 5, 6, 2}, {5, 1, 2, 2}}
+                         };
+                         
+        Pair<double[][], double[]> csp_data = DataFilter.get_csp(data, labels);
```
Andrey1994, line 24: doesnt match code style for java, you can configure IDE(eclipse) to format java code automatically using https://github.com/brainflow-dev/brainflow/blob/master/java-package/brainflow_formatter.xml smth like preferences->java->code style->import formatter

julia-package/brainflow/test/csp.jl
```diff
@@ -0,0 +1,11 @@
+using BrainFlow
```
Andrey1994, line 1: in order to julia to run this test automatically you should add it to https://github.com/brainflow-dev/brainflow/blob/master/julia-package/brainflow/test/runtests.jl

python-package/brainflow/data_filter.py
```diff
@@ -609,6 +621,44 @@ def perform_wavelet_denoising(cls, data: NDArray[Float64], wavelet: str, decompo
...
+        if not (len(labels) == data.shape[0]):
+            raise BrainFlowError('Invalid number of elements in array <labels>', BrainflowExitCodes.INVALID_ARGUMENTS_ERROR.value)
+        
+        n_epochs, n_channels, n_times = data.shape
+
+        temp_data1d = numpy.zeros((int(n_epochs * n_channels * n_times))).astype(numpy.float64)
+        for e in range(n_epochs):
+            for c in range(n_channels):
+                for t in range(n_times):
+                    temp_data1d[e * n_channels * n_times + c * n_times + t] = data[e, c, t]
```
Andrey1994, line 646: I think in slack you mentioned that you get rid off such loop in python, did you forget to push this change?

Andrey1994 (changes requested):
it looks great! I wrote a few comments but its just nitpicking.

Also, I would like to create a demo with real data, current tests more like unit test, lets create a notebook with data from synthetic board and publish it like this https://brainflow.readthedocs.io/en/stable/notebooks/denoising.html

Andrey1994 (approved):

Issue #148, opened 2020-11-29
Add support fort Muse devices


Andrey1994:
We can use BLED112 dongle and BGLIB to add support for Muse devices. Info about BLE services and characteristics can be taken from muse lsl
Use Ganglion Implementation as a referecence

xloem:
Does brainbit-web support the muse?
I've worked with the muse a little.  Here are the gatt characteristics: https://github.com/alexandrebarachant/muse-lsl/blob/master/muselsl/constants.py
The serial characteristic accepts commands that initiate notification streaming on the others and configure which ones to stream.  most commands are used here: https://github.com/alexandrebarachant/muse-lsl/blob/master/muselsl/muse.py
The muse has hexadecimal 'profiles' that enable or disable different characteristics.  here are the profiles for the latest muse S; they are likely a superset of the older profiles: https://github.com/xloem/pymuse/blob/master/muse.py#L52

There are different kinds of muses:
- the original muse which was a BT serial device instead of BTLE (but accepted mostly the same commands)
- the new muses which are all BTLE: the muse 2016, muse 2, and muse S.  these work basically the same as each other but may have different different profile characteristics.

matthijscox:
We are looking for someone with a Muse and c/c++ experience to add it to BrainFlow. 

See the docs for some instructions https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#instructions-to-add-new-boards-to-brainflow.

Andrey can also help you further if you are willing to volunteer ;)

Andrey1994:
"Does brainbit-web support the muse?"
Sorry, looks like I was writing two issues at the same time and copypasted smth wrong. It was about #147

Andrey1994:
I saw muse-lsl and going to use it as a reference, have no device to start working on it, maybe will get one soon

xloem:
I think the original muse-lsl research was borrowed from https://github.com/urish/muse-js .  There's also information at sdk archives like https://sites.google.com/a/interaxon.ca/muse-developer-site/home .  But the information in my first post is basically what's needed.

I might have some time to work on adding muse but I likely won't finish on my own (unless somebody can pay me or something ;p).  Is this repository a place to store partial work?

Andrey1994:
I can create a branch for muse and its ok if its in WIP state.

I wish I could pay for such work, but there is no company\research grant behind this project, so as of right now there is no such opportunity

xloem:
[...]

PR #215, opened 2021-01-24
Opportunistic Muse 2 work

xloem:
Note this PR merges into the `muse` branch, not the main code.

Current status: connects to a nearby muse and attempts to configure it.  the bluetooth c++ driver i found is not very stable; it either needs debugging or replacing.  i'm working on debugging it next.  it would be valuable to look for other bluetooth c++ drivers.


CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 cmake_minimum_required (VERSION 3.13)
 project (brainflow)
 
-set (CMAKE_CXX_STANDARD 11)
+# changed from 11 to 17 for nativeble
+set (CMAKE_CXX_STANDARD 17)
```
Andrey1994, line 5: we can not move to c++17, it will break compilation on devices with old compiler(like raspberry, etc)

CMakeLists.txt
```diff
@@ -320,6 +327,15 @@ if (USE_OPENMP)
     endif (OpenMP_CXX_FOUND)
 endif (USE_OPENMP)
 
+# what library is used for bluetooth is currently decided by a preprocessor define at the top of src/utils/bluetooth.cpp
+target_link_libraries (${BOARD_CONTROLLER_NAME} PRIVATE nativeble)
+
+find_package(PkgConfig REQUIRED)
+pkg_search_module(GATTLIB REQUIRED gattlib)
```
Andrey1994, line 334: am I right that it impicates linux and bluez?

This way we will have to implement it for each operation system since ble stack is very different between OSes.

I thought that we should use BLED112 dongle and use BGLIB(serial port in fact), ganglion implementation can be used as a reference.

Also for bluez related api I have a concern about licenses, most of them distributed under GPL which is not compatible with MIT

csharp-package/brainflow/brainflow/board_controller_library.cs
```diff
@@ -63,7 +63,9 @@ public enum BoardIds
         NOTION_2_BOARD = 14,
         IRONBCI_BOARD = 15,
         GFORCE_PRO_BOARD = 16,
-        FREEEEG32_BOARD = 17
+        FREEEEG32_BOARD = 17,
+        MUSE_1_BOARD = 18,
```
Andrey1994, line 67: I think there is 0 chance that support for both muse devices will be added in single PR, more likely it will be splitted by multiple PRs

src/board_controller/muse/muse2.cpp
```diff
@@ -0,0 +1,399 @@
...
+        // The current error code selections in this sourcefile could be considered stubs if
+        // insufficient.
+        return (int)BrainFlowExitCodes::UNABLE_TO_OPEN_PORT_ERROR;
+    }
+
+    // find and connect to device
+
+    if (params.mac_address.empty ())
+    {
+        std::cerr << "No mac address provided, scanning for a nearby muse ..." << std::endl;
```
Andrey1994, line 43: use logger instead std::cerr

Andrey1994:
If newer standard is needed and if for some reason you dont want to reuse BGLIB approach like for ganglion(I really prefer this way, its x-platform and well known), code which uses C++17 should be moved to another DLL\SO\dylib file and exposed as a dynamic library to the core module.(check oymotion\unicorn\ganglion devices, they load libraries in runtime)

CMakeLists.txt
xloem: thanks for commenting on these things.  i haven't read eveyrthing yet but am taking some time to reply to some.

nativeble uses c++17 for now (although could be downgraded, the author is accepting all my pull requests).  nativeble is just for prototyping, but could be kept and c++17 only enforced if nativeble is enabled

CMakeLists.txt
xloem: these are just libraries I linked in to get bluetooth working.  nativeble is MIT and cross-platform, but not very stable.  gattlib is GPL and linux-only but is just for testing.  I mentioned these on the muse issue thread near the top.

I didn't realise the BLED112 dongle also provides a full bluetooth interface.  I mistakenly thought it was domain-specific like the cyton dongle.

The interface I wrote to these libraries should be mergeable with the ganglion interface, but I would need to purchase a BLED112 dongle to test.

A notable advantage of general-purpose bluetooth is that users would not have to purchase a specific dongle, but could optionally use the existing bluetooth support of their devices, if they install a library that connects with them.

csharp-package/brainflow/brainflow/board_controller_library.cs
xloem: sounds like it's better to leave the board_id out if there's no work to implement it yet

src/board_controller/muse/muse2.cpp
xloem: i'm just trying to get it up and running now, so focusing my time on what makes it run

xloem:
Thanks Andrey, I was unaware of BGLIB.  I thought it was a domain-specific dongle.  Feel free to hack this code up.

CMakeLists.txt
Andrey1994: "A notable advantage of general-purpose bluetooth is that users would not have to purchase a specific dongle, but could optionally use the existing bluetooth support of their devices, if they install a library that connects with them."

thats true, when I was working on ganglion my first try was to use OS specific libs, but later I realised that all of them are completely different and all of them have different issues(example windows native SDK in C++ has some bugs, so it doesnt work with C# gui apps and requires CLR to be compiled)

xloem: given i don't have a bled112 at this time, but am happy to get a bluetooth backend working, any thoughts?  if i migrate the bglib code to a generic interface in src/utils that still uses bglib, would you be interested in testing it with a ganglion?  the problem is that i make a lot of bugs so the test cycle can be very slow.

that would leave some space for adding platform interfaces for people like me without dongles, once the work is put in to make them work.  also opens avenues for mobile devices where dongles are harder but ble is present.

i've also been thinking about labstreaminglayer some.  it seems some work is duplicated between here and labstreaminglayer.

src/board_controller/muse/muse2.cpp
xloem: this seems really low priority to me, given i don't have too much time to add here.  it seems what is needed here is adding more muse bluetooth support, and resolving the bglib/bluetooth.cpp difference.

Andrey1994:
"It seems strange that openbci states the ganglion requires the BLED112 dongle when it is just speaking normal BTLE."

I dont think so, its kinda a tradeoff between deveopers efforts and user friendliness, if you take a look at other devices which use BLE you will find that almost all of them use dongle to work with it(unicorn, oymotion, etc) the only exception here is brainbit and its a reason why their official libs work only for windows. Also users with workstations will need to buy a dongle since there is no ble support for workstations(only for laptops)

Before even for openbci it was diffenret(for macos native api was used and for other OSes dongle was used). Now its uniform and uses dongle for all OSes.

I prefer to have a unform solution even if it forces users to buy addition dongle

xloem:
in python the bleak library makes cross-platform btle easy.  it seems that is not the case for c++ =/

i'm still learning about the ecosystem here.  it seems that EEG device developers are really strapped for software development effort.  where do you think software development effort is needed most?

Andrey1994:
"where do you think software development effort is needed most?"

in this project))

Andrey1994:
example of using bglib https://github.com/brainflow-dev/brainflow/pull/192 its in wip stage and not written by me. For ganglion its almost the same

bglib forces users to implement global callbacks wo namespaces and these callbacks should be different between devices and its really bad. It also means that you can not reuse it for multiple boards as is since there will be multiple definitions/implementations error.

To bypass this error I moved bglib part to its own dll loaded manually in runtime when I was implementing it for ganglion. Also, there are a lot of other dirty hacks to make it work. But it does the job and should be relatively simple to reuse

xloem:
I haven't found the time yet to address everything you've brought up.

> example of using bglib #192 its in wip stage and not written by me. For ganglion its almost the same
> 
> bglib forces users to implement global callbacks wo namespaces and these callbacks should be different between devices and its really bad. It also means that you can not reuse it for multiple boards as is since there will be multiple definitions/implementations error.
> 
> To bypass this error I moved bglib part to its own dll loaded manually in runtime when I was implementing it for ganglion. Also, there are a lot of other dirty hacks to make it work. But it does the job and should be relatively simple to reuse

It sounds like it could still be generalised and put into utils, but as-is you'd need to load a new copy of a dll or do some symbol mangling for each connected device.  Maybe it would be easier to modify the source of bglib a little bit, to pass a user pointer to the callbacks or take a struct that has the callbacks as member variables.

I googled bglib and it seems you need to register an account with silabs and download and install simplicity studio to get it.  A lot of license agreements.  It seems they also have a user-implementable device API.

Andrey1994:
You dont need to install anything for bglib, its already bundled with the code.

I thought to change its code and found its too complicated. There are a few common files like uart, cmddef, etc but what we really need is files for callbacks, stubs and main and they are different device to device

xloem:
Linking this to #148 and closing it because I never ended up finding time to organise it as requested; I think andrey is planning a different approach for the Bluetooth implementation, but this PR is still here and theoretically reasonable if not.

PR #221, opened 2021-02-08
refactor osserial and libftdiserial

Andrey1994:


src/utils/libftdi_serial.cpp
```diff
@@ -0,0 +1,252 @@
...
+        return (int)SerialExitCodes::CLOSE_ERROR;
+    }
+}
+
+const char *LibFTDISerial::get_port_name ()
+{
+    return description.c_str ();
+}
+
+#else
```
xloem, line 192: I might be missing something, but it looks to me like this block is unneeded if the libftdi interface is kept private, because the only way to access it is guarded by the same check: https://github.com/brainflow-dev/brainflow/pull/221/files#diff-707d50ff0c1f53ceed4b00bbb44de221f4e4804adeb6c8a62059677a2f153933L443

xloem:
Thanks for implementing this.  I hadn't gotten around to it yet myself.  It looks fine to me.  The choices are all organisational.

src/utils/libftdi_serial.cpp
Andrey1994: no, because now there is a header for libftdi class and this header included wo ifdef. Can change it and include header under ifdef check also, I like this way because there is no need to write ifdefs outside class implementation. But its just a preference, both options are fine

PR #223, opened 2021-02-08
Send to board

GrahamBriggs:
Fixed the send_to_board function, and fixed the "Failure" string compare.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -372,3 +372,5 @@ Makefile
 # Libraries
 *.so
 *.a
+CMakeSettings.json
+._.gitignore
--- a/csharp-package/brainflow/brainflow/brainflowNetStandard.csproj
+++ b/csharp-package/brainflow/brainflow/brainflowNetStandard.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <AssemblyName>brainflow</AssemblyName>
     <RootNamespace>brainflow</RootNamespace>
   </PropertyGroup>
--- a/csharp-package/brainflow/brainflow/brainflowNetStandard.csproj
+++ b/csharp-package/brainflow/brainflow/brainflowNetStandard.csproj
@@ -1,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
+    <AssemblyName>brainflow</AssemblyName>
+    <RootNamespace>brainflow</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Remove="Properties\**" />
+    <EmbeddedResource Remove="Properties\**" />
+    <None Remove="Properties\**" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="brainflow.nuspec" />
+    <None Remove="brainflow.targets" />
+    <None Remove="packages.config" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Accord.Math" Version="3.8.0" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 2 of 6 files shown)


Andrey1994:
this action shows what should be changed to match code style using git diff https://github.com/brainflow-dev/brainflow/pull/223/checks?check_run_id=1857559352

Why do we need netstandard file? And I see net472 there, will it work for other .net versions? I am not an expert in C#...

GrahamBriggs:
Andrey

Thank you for reviewing my work. I apologize because I did not know that pushing additional changes to that branch would trigger more pull request review and tests.

I reviewed the report for code style. I can not see what it is changing, my guess is spaces/tabs?  Is there any way to accept the reformatted.  I am not keen to install a new code formatting tool in my IDE.

Regarding need for .NET standard project.  Part of my system includes a cross platform mobile application built using Xamarin forms.  This requires .NET standard version of the assemblies.


Regarding the net472 in the <TargetFrameworks> element:

This would normally not be required. NetFramework assemblies and apps can normally consume NetStandard assemblies without the need to specify the version of .NET framework in the NetStandard assembly project.  However, there is a bug in the Accord math nuget package.  The code will build, but if you run it you get this:


The solution to this problem is to explicitly define your net framework assemblies in the net standard project file.  See this post:  https://github.com/accord-net/framework/issues/1426 <https://github.com/accord-net/framework/issues/1426> 

You need to do this for all assemblies that might interact with accord or that interact with any other assembly that is interacting with Accord.

Full disclosure:  I am not much of a C# expert either, I muddle my way through this kind of detail when I need to, but I am much more comfortable when this kind of stuff just works and I can spend my time making the software do useful things (for example I wasted three to four hours fighting the Accord math target framework issue before discovering the fix).  I believe if you were targeting older version of .NET, you would also have to add that in the target frameworks. I am not familiar enough to know if older versions of the .NET framework support NetStandard assemblies, so maybe this is not an issue?

Thank you again for your time and attention.


GB
=-=-=-=-=-=-=
Graham Briggs <mailto:GrahamBriggs@littlebytesofpi.com>
LittleBytesOfPi.com <http://littlebytesofpi.com/>

GrahamBriggs:
Andrey

I did some more experimentation, removing the net472 from all the Net Standard assembly project files. The program crashes as expected.

However, I then tried to remove this spec one assembly at a time.  Turns out that this is only required in some projects and not others. I can’t explain why.  However the good news is that all four of my C# solutions (three NetFramework and one NetStandard) work with brainflow assembly when the net472 is not there.

I have checked in an updated brainflowNetStandard.csproj that includes only the netstandard20 framework specified. I believe this will be more compatible for other users that may want to use the NetStandard version of the assembly.

Thank you

GB
=-=-=-=-=-=-=
Graham Briggs <mailto:GrahamBriggs@littlebytesofpi.com>
LittleBytesOfPi.com <http://littlebytesofpi.com/>

Andrey1994:
"there is a bug in the Accord math nuget package"

I am going to get rid of Accord as a dependency soon.
So, probably it will fix your issue, lets remove netstandard file from this PR right now, since I would like to test it by myself first and it can take some time(after all it's not related to sending commands to openbci boards). Maybe later, after testing and when I finally understand how it works and why its needed we will restore it

There is no way to accept this diff and it doesnt make changes in your PR automatically(it was informed decision)
At blank lines there are spaces which clang-format removes

image

And there are two lines where it inserts space before bracket, should not be difficult to fix it manually

GrahamBriggs:
Andrey

Thank you, I understand the results of the clang format report better now. I will attempt to fix this manually. I also understand the importance of code consistency so if I ever do additional changes in the library, I will suck it up and invest the time to learn how to use the clang formatter tool in visual studio.

I did see the issue in the tracker about removing accord.  This would be a good step.

I piushed a change to the brainflowNetStandard project: it now only references netstandard2.0 as a target framework.  I was able to scrub the net4.7 out of one additional assembly that references brainflow, but does not invoke any matrix stuff. I have had to retain the net4.7 as a target framework for any of my assemblies that reference brainflow and also invoke Accord methods. I can’t really explain why this is, but it sort of makes sense based on what that post said about Accord nuget package, and it works, so moving on.

FWIW:  Using the code checked in at commit e42d <https://github.com/brainflow-dev/brainflow/pull/223/commits/7e1e42d6273c184896ac09e6e0a4ca1bebc4e05d>  I have successfully built, run, and debugged my C# Xamarin app on Android, my C# WPF app on Windows 64, and my C# Winform app on both Windows 64 and Raspberry Pi (using Mono). I also have a C++ app that runs on the Raspberry Pi and I have tested it with the build of the brainflow libs as well.

I will continue to test and also fix up the formatting of the code on my branch before I push it again.

Thank you for your time and attention. 


GB
=-=-=-=-=-=-=
Graham Briggs <mailto:GrahamBriggs@littlebytesofpi.com>
LittleBytesOfPi.com <http://littlebytesofpi.com/>

PR #226, opened 2021-02-10
Github Action: Consolidate android architectures into a single for loop

xloem:
Reducing the code duplication makes it easier to work with the github workflows, for me.

### DIFF
--- a/.github/workflows/run_android.yml
+++ b/.github/workflows/run_android.yml
@@ -29,30 +29,14 @@ jobs:
       run: |
         echo "y" | sudo ${ANDROID_HOME}/tools/bin/sdkmanager --install "ndk;21.0.6113669" --sdk_root=${ANDROID_SDK_ROOT} > /dev/null
     # build for all ABIs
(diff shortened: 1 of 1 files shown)


Andrey1994:
I think if compilation fails for ABIs 1-3 and doesnt fail for the latest one entire job will be marked as "Successfull" because exit code will be 0

xloem:
@Andrey1994 I reviewed the behavior and it does correctly fail if any architecture fails.

Github workflows [use `set -eo pipefail`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference) to fail if any individual line fails.

I tested this by [simulating a failure for x86](https://github.com/xloem/brainflow/pull/4/commits/2c797dc0d4a3f4326b34d19f5e8674f8db2780b3) and [the workflow correctly failed](https://github.com/xloem/brainflow/pull/4/checks?check_run_id=1871939228#step:6:256).

PR #228, opened 2021-02-10
libftdi: handle ftdi_new failing

xloem:
Adds code to handle the edge case of ftdi_new returning 0, which happens when libusb_init fails.


xloem:
Note: I'm learning the insides of libftdi a little, and it looks like a cleaner solution to this would be to use ftdi_init/ftdi_deinit instead of ftdi_new/ftdi_free.  This preserves libftdi's internal error messages when initialisation fails, and also saves a minor heap allocation.

Andrey1994:
Since I already merged it lets do it in another PR

Issue #142, opened 2020-11-24
Remove accord as a dependency from C# binding


Andrey1994:
Accord is not supported by it's author anymore, we can get rid of it and use plain arrays

Andrey1994:
done in new_major branch

PR #232, opened 2021-02-13
rm accord as a dependency

Andrey1994:

#142 

Other options I considered:

* keep accord, the only problem with it: a little more complicated setup process and it's archived
* use nd arrays as is, but there are no methods to access rows\cols. Row access is needed for signal processing methods and for users to get data for entire channel
* use jagged arrays, row access will work from the box but it will be a breaking change  and not easy to work with
* write own class for NDArrays and probably do the same for other bindings
* use https://github.com/SciSharp/NumSharp 

How its implemented here:

* accord removed
* it's a multidemsional array but I ported very small subset of accord extensions to work with matrixes directly into C# binding. Directly in brainflow we used just a few methods from Accord, and they should work as is right now even wo accord. But it still can break users code if they call other methods from Accord

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -134,8 +134,6 @@ jobs:
         VERSION: ${{ github.event.inputs.version }}
     - name: Prepare Nuget Package
       run: |
-        nuget install Accord -OutputDirectory $env:GITHUB_WORKSPACE\csharp-package\brainflow\packages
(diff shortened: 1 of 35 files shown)


xloem (approved):
I'm not familiar with c-sharp at this time, but removing unmaintained dependencies sounds pretty helpful.  I see a lot of files were trimmed, and a new one that looks well-made was added to fill the gap.  It might be helpful to express interest in the sourcefile or documentation, around integrating any mainstream matrix library people users use and appreciate if one arises.

PR #233, opened 2021-02-13
Upgrade JNA

xloem:
While working with android I noticed the JNA version used in brainflow is pretty old.

Here's a small PR to upgrade that version.


xloem:
@Andrey1994 I've been otherwise occupied but I wanted to make sure you knew that PR #233 broke android support because I did not upgrade the jnidispatch.so files as well.  Maybe you've bumped into that and fixed it, but if not either #233 should be reverted, or the jnidispatch.so files also upgraded.  I had reverted this I think in the maven PR, but we never ended up resolving that.

PR #237, opened 2021-02-20
libftdi: init with member storage to retain init errors

xloem:
I mentioned this change some time ago; storing the libftdi object as a class member provides for logging initialisation errors and saves a heap allocation.


xloem:
I'll review these more since they are displaying intermittent test failures on macos now apparently.

PR #242, opened 2021-02-22
add platform_ptr to params, and pass JNIEnv pointer in for java

xloem:
@Andrey1994 I'm curious how you'd like this to look, or if this approach is acceptable.

This was a quick change I made to the insides of the library to make accessing android much easier, which requires the JNIEnv pointer, but I don't really know the norms of the board params structure very well.  Is this how you would have done it?

Adding the pointer to params and adding a getter for params means that anything with a board pointer can get the platform pointer, which on java is now the JNIEnv pointer providing access to the Java VM.

I haven't added documentation or anything for this pointer, but it's presently how I'm making android work.


Andrey1994:
BrainFlowInputParams is definitely the wrong place to add it, its for user proivided info.
Also, instead default argument in prepare_session I prefer another overload of prepare_session method or maybe make it under ifdef.

Also, I guess such stuff should be merged in the same PR which uses this ptr

xloem:
Thanks so much for your feedback.

PR #243, opened 2021-02-25
add brainflow array class to cpp binding

Andrey1994:

### DIFF
--- a/cpp-package/src/data_filter.cpp
+++ b/cpp-package/src/data_filter.cpp
@@ -77,7 +77,7 @@ double *DataFilter::perform_downsampling (
 }
 
 std::pair<double *, int *> DataFilter::perform_wavelet_transform (
-    double *data, int data_len, char *wavelet, int decomposition_level)
+    double *data, int data_len, std::string wavelet, int decomposition_level)
 {
     if (data_len <= 0)
     {
@@ -88,8 +88,8 @@ std::pair<double *, int *> DataFilter::perform_wavelet_transform (
     double *wavelet_output = new double[data_len +
         2 * decomposition_level * (40 + 1)]; // I get this formula from wavelib sources
     int *decomposition_lengths = new int[decomposition_level + 1];
-    int res = ::perform_wavelet_transform (
-        data, data_len, wavelet, decomposition_level, wavelet_output, decomposition_lengths);
+    int res = ::perform_wavelet_transform (data, data_len, const_cast<char *> (wavelet.c_str ()),
+        decomposition_level, wavelet_output, decomposition_lengths);
     if (res != (int)BrainFlowExitCodes::STATUS_OK)
     {
         delete[] wavelet_output;
@@ -100,7 +100,7 @@ std::pair<double *, int *> DataFilter::perform_wavelet_transform (
 }
 
 double *DataFilter::perform_inverse_wavelet_transform (std::pair<double *, int *> wavelet_output,
-    int original_data_len, char *wavelet, int decomposition_level)
+    int original_data_len, std::string wavelet, int decomposition_level)
 {
     if (original_data_len <= 0)
     {
@@ -109,8 +109,9 @@ double *DataFilter::perform_inverse_wavelet_transform (std::pair<double *, int *
     }
 
     double *original_data = new double[original_data_len];
-    int res = ::perform_inverse_wavelet_transform (wavelet_output.first, original_data_len, wavelet,
-        decomposition_level, wavelet_output.second, original_data);
+    int res = ::perform_inverse_wavelet_transform (wavelet_output.first, original_data_len,
+        const_cast<char *> (wavelet.c_str ()), decomposition_level, wavelet_output.second,
+        original_data);
     if (res != (int)BrainFlowExitCodes::STATUS_OK)
     {
         delete[] original_data;
@@ -120,9 +121,10 @@ double *DataFilter::perform_inverse_wavelet_transform (std::pair<double *, int *
 }
 
 void DataFilter::perform_wavelet_denoising (
-    double *data, int data_len, char *wavelet, int decomposition_level)
+    double *data, int data_len, std::string wavelet, int decomposition_level)
 {
-    int res = ::perform_wavelet_denoising (data, data_len, wavelet, decomposition_level);
+    int res = ::perform_wavelet_denoising (
+        data, data_len, const_cast<char *> (wavelet.c_str ()), decomposition_level);
     if (res != (int)BrainFlowExitCodes::STATUS_OK)
     {
         throw BrainFlowException ("failed to perform wavelet denoising", res);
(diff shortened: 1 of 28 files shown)


xloem:
Andrey thanks for the review request ... it's hard to sift through the formatting changes all over the code to find the impactful ones; where do I look to see the bulk of the work?

Andrey1994:
I will check valgrind issue also, have not run it locally yet.
There is only one minor code style change in templates, most of the changes about using c++ objects instead raw pointers. 

The most important change here is BrainFlowArray class, its inspired by smth like numpy. Idea is to store all data from ndarray in continuous memory chunk. https://github.com/brainflow-dev/brainflow/blob/1f49cb828a95f44ddfffcebc5b237acb8e0b35bc/cpp-package/src/inc/brainflow_array.h

Andrey1994:
And I am thinking about adding such ndarray implementation for some other languages, it should increase perf and avoid reshaping/memcpy for some algorithms. 

Problem which it should solve:

you can pass only 1d array from java\python\c#\etc to c++. But for some methods we need to pass 2d\3d arrays, it means that we need to reshape and copy data each time we call them.  If we represent array like here we dont need to reshape because its continuous from the box

Andrey1994:
valgrind issue looks like an error in move constructor in BrainFlow array, should use memove instead memcpy and free data array

xloem:
no need to memcpy on move constructor, you can just reassign the pointer

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+
+    return result;
+}
+
+template <size_t N>
+static inline std::array<int, N> zero_array ()
+{
+    std::array<int, N> result;
+
+    for (int i = 0; i < N; i++)
```
xloem, line 37: If you wanted, you could use [`result.fill(0)`](https://en.cppreference.com/w/cpp/container/array/fill) here instead of a for loop.  This would let the standard library use an implementation optimised for the platform.

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+    {
+        result[i] = 0;
+    }
+
+    return std::move (result);
```
xloem, line 42: Usually one does not use std::move for returning values due to [return value optimization](https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement).

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+}
+
+static inline std::array<int, 1> make_array (int size0)
```
xloem, line 45: The parameters here are named `size`, but the resulting arrays are holding them as values, not sizes.  Not sure if that is intentional.  These are consolidatable into one template function, not sure if it is your style to do so:
```
template <typename... types>
static inline std::array<int, sizeof...(types)> make_array (types ... values)
{
    return std::array<int, sizeof...(types)>{values...};
}
```

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+{
+
+private:
+    int length;
+    std::array<int, Dim> size;
+    std::array<int, Dim> stride;
+    T *origin;
+
+public:
+    friend std::ostream &operator<<<> (std::ostream &out, const BrainFlowArray<T, Dim> &array);
```
xloem, line 95: this is new to me .. what is `operator<<<>` ?

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    BrainFlowArray (BrainFlowArray &&other)
+        : length (other.length)
+        , size (std::move (other.size))
+        , stride (std::move (other.stride))
+        , origin (nullptr)
+    {
+        origin = new T[other.length];
+        memcpy (origin, other.origin, sizeof (T) * other.length);
```
xloem, line 211: Here there is no need for a memcpy.  You can simply say
```
origin = other.origin;
other.origin = nullptr;
```

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+        other.origin = nullptr;
+    }
+
+    const BrainFlowArray &operator= (BrainFlowArray &&other)
+    {
+        length = other.length;
+        size = std::move (other.size);
+        stride = std::move (other.stride);
+        origin = new T[other.length];
+        memcpy (origin, other.origin, sizeof (T) * other.length);
```
xloem, line 224: similarly here.
```
if (origin != nullptr)
{
    delete [] origin;
}
origin = other.origin
other.origin = nullptr;
```

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+        return get_raw_ptr ();
+    }
+
+    operator const T * () const
+    {
+        return get_raw_ptr ();
+    }
+
+    /// access element at index
+    T &at (int index0)
```
xloem, line 301: C++ developers usually distinguish between `at()` and `operator[]` in that `at()` throws an out of bounds exception whereas `operator[]` does not check bounds.  Here you do not throw an exception which would be misleading to me, but I don't know whether the same paradigm holds true for researchers as opposed to developers.

cpp-package/src/inc/brainflow_array.h
```diff
@@ -0,0 +1,468 @@
...
+
+    /// fill already preallocated buffer
+    void fill (T *ptr, int size)
+    {
+        memcpy (origin, ptr, sizeof (T) * size);
+    }
+};
+
+template <typename T, size_t Dim>
+std::ostream &operator<< (std::ostream &os, const BrainFlowArray<T, Dim> &array)
```
xloem, line 428: thanks for implementing this, people don't always add a way to quickly output vectors and amtrics, and it really helps

xloem (changes requested):
This looks fine.  The assignment operator should call `delete [] origin` to prevent memory leak.  I added many line comments but they are not necessary.  But the assignment operator should call `delete [] origin` before assignment to it, to free the memory.  A similar addition is needed to the move constructor if the memcpy there is kept, but there is no need for a memcpy in a move constructor.

[edit: I was thinking about this and noticed it's in the cpp bindings.  If it were in src/utils then all the board drivers could use it, and it could be used as an internal data format.  regarding reshaping, it is possible to reorder the dimensions without doing any memcpys: it's notable however that other libraries have already implemented that]

cpp-package/src/inc/brainflow_array.h
xloem: If you do keep the memcpy, you'll want to `delete [] other.origin` here to prevent a memory leak.

Andrey1994:
For all boards there is a ringbuffer which stores data and its different(https://github.com/brainflow-dev/brainflow/blob/master/src/utils/data_buffer.cpp) we can always move to utils latern on if it will be needed

cpp-package/src/data_filter.cpp
```diff
@@ -151,7 +151,7 @@ std::pair<BrainFlowArray<double, 2>, BrainFlowArray<double, 1>> DataFilter::get_
         throw BrainFlowException ("failed to compute the CSP filters", res);
     }
 
-    return std::make_pair (std::move (filters), std::move (output_eigenvalues));
+    return std::make_pair (filters, output_eigenvalues);
```
xloem, line 154: just noting that here `std::move` does indeed have an effect in this case, because the values are passed to `std::make_pair` instead of returned directly.

PR #249, opened 2021-02-26
Android LibFTDI Serial USB Support for Cyton

### LINKED ISSUE #216: Cyton on Android
I had success accessing the Cyton dongle on android using the python jni driver at https://github.com/jacklinquan/usbserial4a .  Unfortunately, that driver is python rather than c/c++ .

> Here is the driver code used by usbserial4a:
> - https://github.com/jacklinquan/usb4a/blob/master/usb4a/usb.py
> - https://github.com/jacklinquan/usbserial4a/blob/master/usbserial4a/ftdiserial4a.py
>
> usbserial4a is making extensive use of https://github.com/kivy/pyjnius , a python library that transparently proxies access to java objects with the jni.

If the approach of usbserial4a is the only one that works, then the solution may involve an FTDI driver using android's USB-OTG API, accessed via JNI.

Here are possible resources:
- libusb has android support.  I haven't tested it.  It looks like it requires use of the JNI to enable permission.  I opened an issue here: https://github.com/libusb/libusb/issues/862  UPDATE: in a comment to that issue, somebody explained that you can use the JNI to get the usb device filedescriptor, and pass it to libusb
- libftdi is able to access the cyton dongle via libusb https://www.intra2net.com/en/developer/libftdi/ 
- https://github.com/SpecLad/libusb-android an outdated port of libusb to android
- https://github.com/chzhong/serial-android a port of a serial library to android, not sure how to build it
- https://github.com/bytedeco/javacpp lists a number of resources for moving between java and c++ in its readme: [Jace](https://bitbucket.org/cowwoc/jace/), [JunC++ion](http://codemesh.com/products/junction/), [JCC](http://lucene.apache.org/pylucene/jcc/), [jni.hpp](https://github.com/mapbox/jni.hpp), [Scapix](https://www.scapix.com/)

xloem:
This connects the serial backend to https://github.com/libusb/libusb/pull/874 to get usb permissions, so that usb ftdi chips can be opened on android.

- [ ] retest reading from board on android

I've tested and worked with this and it's running with my test app (which is currently a small pile of hacks on top of [BrainFlowAndroidTest](https://github.com/brainflow-dev/BrainFlowAndroidTest)) on my android phone again.  The force pushes below were updates and fixes.

After this PR it would make sense to make one to BrainFlowAndroidTest to add interfacing for this.  That work might inform design changes here, too.

Meshes with https://github.com/brainflow-dev/brainflow/pull/251 which I made to help test it.
Fixes #216 .
Completes #247 [which got derailed by some of my possibly-delusional paranoia, sorry about that].


.github/workflows/run_android_libftdi.yml
```diff
@@ -0,0 +1,74 @@
+name: Run libFTDI Android NDK
```
Andrey1994, line 1: if I am not mistaken wo libftdi boards which use serial port do not work, I think we can keep it in the same workflow

CMakeLists.txt
```diff
@@ -530,6 +530,17 @@ if (ANDROID)
     add_custom_command (TARGET ${BOARD_CONTROLLER_NAME} POST_BUILD
         COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${CMAKE_HOME_DIRECTORY}/compiled/${BOARD_CONTROLLER_COMPILED_NAME}" "${CMAKE_HOME_DIRECTORY}/tools/jniLibs/${ANDROID_ABI}/${BOARD_CONTROLLER_COMPILED_NAME}"
     )
+    foreach (DEPENDENCY_LIBRARY ${LIBFTDI_LIBRARIES})
```
Andrey1994, line 533: should it be under if (libftdi enabled)? and what is the content of this variable? Maybe its shorter wo loop?

src/board_controller/inc/board_controller.h
```diff
@@ -38,6 +38,10 @@ extern "C"
     SHARED_EXPORT int CALLING_CONVENTION set_log_file (char *log_file);
     SHARED_EXPORT int CALLING_CONVENTION log_message (int log_level, char *message);
 
+    // platform methods
+    SHARED_EXPORT int CALLING_CONVENTION java_set_jnienv (
+        const struct JNINativeInterface *java_jnienv);
```
Andrey1994, line 43: I can not find a header for it in includes.. how does it work?

CMakeLists.txt
xloem: `LIBFTDI_LIBRARIES` is filled with all libraries libftdi needs linked, when libftdi is found.  It's empty (undefined values are empty) otherwise.  I did it in this format so that if there are ever other dependency libraries, they can be copied in simply by adding to the loop.  The loop should be kept in case libftdi changes what libraries it uses (like if libusb-1.0 becomes libusb-1.1).  Do you think there's value to wrapping it in an `if`?

src/board_controller/inc/board_controller.h
xloem: Android and java backends often use [jni.h](https://android.googlesource.com/platform/libnativehelper/+/brillo-m9-dev/include/nativehelper/jni.h) to interoperate with java.  It's been standard for some time, but is only present in java environments.  To perform any java calls using jni.h, you need a `JNIEnv` pointer, which is a shorthand waying of saying a `struct JNINativeInterface` pointer.  The java runtime hands this pointer to all native methods.  Here I used an [elaborated type specifier](https://en.cppreference.com/w/cpp/language/elaborated_type_specifier) to declare `struct JNINative Interface` at the same time as using it in a pointer, which means jni.h is not needed to build the library.

I liked the previous approach of using a `void * platform_ptr` which could be used for other things on non-java platforms, but since you asked to only do java I thought it would be clearer to specify the type exactly.

.github/workflows/run_android.yml
```diff
@@ -32,12 +32,29 @@ jobs:
       run: |
         echo "y" | sudo ${ANDROID_HOME}/tools/bin/sdkmanager --install "ndk;${ANDROID_NDK_VERSION}" --sdk_root=${ANDROID_SDK_ROOT} > /dev/null
     # build for all ABIs
+    - name: Compile libusb
+      run: |
+        git clone --depth 1 --branch android https://github.com/xloem/libusb.git
```
Andrey1994, line 37: I saw your PR in upstream, there is very good feedback. How do you think when they will merge it?

src/utils/inc/serial.h
```diff
@@ -10,7 +10,8 @@ enum SerialExitCodes : int
     SET_TIMEOUT_ERROR = -4,
     CLOSE_ERROR = -5,
     NO_SYSTEM_HEADERS_FOUND_ERROR = -6,
-    NO_LIBFTDI_ERROR = -7
+    WRONG_DRIVER_FOR_PORT = -7,
```
Andrey1994, line 13: lets keep naming similar, add _ERROR in the end

src/board_controller/inc/board.h
```diff
@@ -22,6 +22,7 @@ class Board
 {
 public:
     static std::shared_ptr<spdlog::logger> board_logger;
+    static JNIEnv *java_jnienv;
```
Andrey1994, line 25: I already asked about it but still dont fully understand it. I suppose somewhere file like jni.h should be included in order for it to work. Specially since its not under smth like ifdef android. But I dont see it in headers

xloem: I added a declaration here to make it clearer:
https://github.com/brainflow-dev/brainflow/blob/eca05e8b714e23d7847bb02ba1b41aedb64f5162/src/board_controller/inc/board_controller.h#L7
I simply gave it the same name as in the jni.h header, so that if the jni library is used, the type is correct.  Because it is just a pointer the type only need be declared, not defined.  It does this explain it?

JNI can be used on any java platform, not just android.  Given this, do you think it would still be good to only enable the code on android?  That would ensure jni.h is available.  Many people find extra #ifdefs to make code harder to read, but many do it as a norm too.

src/utils/libftdi_serial.cpp
```diff
@@ -72,17 +58,34 @@ void LibFTDISerial::log_error (const char *action, const char *message)
 
 int LibFTDISerial::open_serial_port ()
 {
+    if (!lib_init)
+    {
+        // failed to init libftdi; at least check the port name
+        if (description.size () == 0 || description[0] == '/')
```
xloem, line 64: Note to self: the indexing check should be protected here; move it to the next if block

.github/workflows/run_android.yml
xloem: Right now it looks like they're looking for successful test reports to move things forward, and that more of those will happen if I add a requested feature.  I was hoping to get the brainflow test app going with the changes and kind of meet two goals with one task, but I haven't gotten to that yet.

src/board_controller/inc/board.h
Andrey1994: yep, now I get it

PR #251, opened 2021-02-28
Add build config for android AAR library.

xloem:
- [ ] review issues with building android using the same pom.xml as desktop
  - [ ] packaging='aar' can't be used to automatically do this anymore, so the zip would be made manually like jna does, unless packaging can be set within a profile
  - [ ] jna transient dependency issue means detecting the usage of the dependency within a profile
    - note: [dependencies are referenceable as properties](https://stackoverflow.com/a/6934552/129550) but possibly mostly just within the same pom file
  - [ ] test until all issues enumeated/solved
- [ ] address remaining PR comments

This is a prototype gradle setup to bundle brainflow into an android library.  It directly references the jna library, so the libjnidispatch.so files have been removed, since maven downloads them.  If that's a concern at all it is easy to change back.

I made this change to help manage version differences between parts.  Gradle and maven take care of all the downloading and comparing now.

usage (with android libs built already into tools/jniLibs):
```
cd java-package/brainflow
mvn install
cd ../.. # EDIT: the rest can now be cd ../android; mvn install
cd android-package
./gradlew publishToMavenLocal
```
This compiles and installs the aar into the local user's maven folder.

Then in `app/build.gradle` of an android application:
```
repositories {
    mavenLocal()
}

dependencies {
    implementation 'org.brainflow:android:0.0.1-SNAPSHOT'
}
```
and none of the jars or .so files need to be manually placed into the application.

I'm new to gradle so this might not be perfectly done, but it still seems pretty helpful.


android-package/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+        repositories {
+            maven {
+                url "$buildDir/repo"
+            }
+        }
+    }
+}
+
+android {
+    compileSdkVersion 30
```
Andrey1994, line 35: if I am not mistaken this one is pretty new, I suppose they are backward compatible, so does it make sense to set it to smth old?

android-package/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+}
+
+dependencies {
+    api ('org.brainflow:brainflow:0.0.1-SNAPSHOT') {
```
Andrey1994, line 39: in CI pipelines we change version by regexp/string replace (I dont like regexp, but maven can not read from env variables)

Andrey1994 (commented):
It looks like we can keep both options aar package and jnilibs, also lets change UI pipeline to push aar package.

And I would like to check if its possible to generate aar wo gradle using maven only, and looks like its possible https://stackoverflow.com/questions/25077131/create-android-archive-libraryaar-with-maven-only I would like to use single build system if it works

android-package/build.gradle
xloem: 23 sound good?

android-package/build.gradle
xloem: > And I would like to check if its possible to generate aar wo gradle using maven only, and looks like its possible https://stackoverflow.com/questions/25077131/create-android-archive-libraryaar-with-maven-only I would like to use single build system if it works

It's possible to do this but takes some working with the maven internals that I haven't figured out yet.  I actually talked with a jna developer and they seemed to imply that using gradle would be the most stable, because google wasn't maintaining the maven path: https://groups.google.com/g/jna-users/c/QUFT16yJyt0/m/gapq1WL9AQAJ

xloem:
> It looks like we can keep both options aar package and jnilibs, also lets change UI pipeline to push aar package.
> 
> And I would like to check if its possible to generate aar wo gradle using maven only, and looks like its possible https://stackoverflow.com/questions/25077131/create-android-archive-libraryaar-with-maven-only I would like to use single build system if it works

It's possible to do this but takes some working with the maven internals that I haven't figured out yet. I actually talked with a jna developer and they seemed to imply that using gradle would be the most stable, because google wasn't maintaining the maven path: https://groups.google.com/g/jna-users/c/QUFT16yJyt0/m/gapq1WL9AQAJ [reply moved from accidental line comment]

android-package/build.gradle
xloem: something maven apparently can read from is -D flags passed on the command line, it seems; not sure if it is easy to use that

xloem:
@Andrey1994 I got it building with maven by pulling out the plugin that was throwing the exception.  I added lines to the deploy workflow to publish it like the jars, but I'm not sure how to test this; how do you test the deployment code?  Also, are the java packages published to maven?

java-package/android/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        <configuration>
+            <ndkArchitecture>arm64-v8a armeabi-v7a x86 x86_64</ndkArchitecture>
+        </configuration>
+      </plugin>
+    </plugins>
+
+  </build>
+  <dependencies>
+    <dependency>
+      <groupId>org.brainflow</groupId>
```
Andrey1994, line 60: it looks a little strange, why do we need brainflow as a dependency inside brainflow pom.xml(I know that its pom.xml for android but still strange)?

.github/workflows/deploy.yml
```diff
@@ -80,6 +80,12 @@ jobs:
     - name: Install 7Zip PowerShell Module
       shell: powershell
       run: Install-Module 7Zip4PowerShell -Force -Verbose
+    - name: Download Android Libs
```
Andrey1994, line 83: I think it should be done inside run_android.yaml workflow. And no need to download anything here

Andrey1994:
"Also, are the java packages published to maven?"

to maven central - not. There is github maven package but in order for this to work you need to provide github token in settings.xml for maven(https://github.com/brainflow-dev/brainflow/packages/450100)

"I added lines to the deploy workflow to publish it like the jars, but I'm not sure how to test this"

Usually I trigger it manually, make change and repeat...I think that aar package can be created and uploaded in run_android.yaml workflow and we can upload it to aws as everything else(also you will be able to check it). In workflow for deploy we can publish it to maven later on

java-package/android/pom.xml
xloem: I'm spending a lot of time looking into options for changing this.  It seems an easier approach to having this not look strange might be to rename the brainflow non-android package to something like 'brainflow-core'.  I am thinking maybe modules but I am very new to maven.  The big issue here is that the maven android plugin does not enforce that transient android dependencies are selected, so if the aar and jar are bundled together in one pom.xml, the wrong jna library is used on android in almost every way of doing it.  profile variables may give access to the entire dependency tree via property reflection, providing a possible avenue for working around this, but it doesn't look very nice and easy.  another approach might be submitting a bugfix to the android maven plugin around the dependency issue, and/or adding a feature to maven to make it easy to depend on parent dependency type; i'm not sure if that would mean requiring users to use a forked plugin version in their android projects or not ... most of the alternatives i'm finding here look kinda large unless we just rename the brainflow dependency or add a comment explaining the situation.  any thoughts?

xloem:
> "I added lines to the deploy workflow to publish it like the jars, but I'm not sure how to test this"
> 
> Usually I trigger it manually, make change and repeat...I think that aar package can be created and uploaded in run_android.yaml workflow and we can upload it to aws as everything else(also you will be able to check it). In workflow for deploy we can publish it to maven later on

I don't believe I have privileges with this project to run that workflow, so you know.

java-package/android/pom.xml
Andrey1994: but can it be splitted by 2 different pom.xmls like pom.xml for PC and android-pom.xml for android? This way aar and jar will not be bundled together and may have the same name.

Andrey1994: And I dont get why in this case you need plain brainflow as a dependency for brainflow android when you can just rebuild it. Looks like I am missing smth here

xloem: > And I dont get why in this case you need plain brainflow as a dependency for brainflow android when you can just rebuild it. Looks like I am missing smth here

I ran into some [bugs i linked to earlier](https://groups.google.com/g/jna-users/c/QUFT16yJyt0/m/gapq1WL9AQAJ) between `maven-compiler-plugin` and `maven-android-plugin`.  After your replies, I spent some more time with this and found ways of handling them (I was wrong; there is actually a new release where some of the issues are fixed; maven-compiler-plugin 3.8.1).  So, this is looking somewhat more possible.

> but can it be splitted by 2 different pom.xmls like pom.xml for PC and android-pom.xml for android? This way aar and jar will not be bundled together and may have the same name.

If they have the same name but are separate pom files, then if they are both installed to the same (even local) maven repository, the later-installed pom.xml overwrites the earlier-installed pom.xml .  For this situation, it might make more sense to either give them separate names, or merge the pom files into one (which may be doable without profile tricks .. might look at that more now that sources can be built under android).

Andrey1994: we can change name for aar package, its ok

Issue #248, opened 2021-02-26
check memory layout for numpy arrays


Andrey1994:
Signal processing API requires row major layout, after transpose memory layout can be changed and it will not work. We can not easily fix it, but we can throw an exception and notify user.

https://blog.paperspace.com/numpy-optimization-internals-strides-reshape-transpose/
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html

PR #253, opened 2021-03-03
add check for expected memory layout in numpy

Andrey1994:

https://github.com/brainflow-dev/brainflow/issues/248


python-package/brainflow/utils.py
```diff
@@ -5,8 +5,6 @@
 
 
 def check_memory_layout_row_major(data):
-    if not data:
```
xloem: i notice you removed this to fix ci.  another approach might be something like `if data is None or len(data) == 0:`.  i'm not as familiar with python as c++ to know what is normal.

tests/python/hardware_test.py
```diff
@@ -1,174 +0,0 @@
-import argparse
```
xloem, line 1: did you mean to delete this file?  it's not mentioned in your change summaries.

python-package/brainflow/data_filter.py
```diff
@@ -383,6 +384,7 @@ def perform_lowpass(cls, data: NDArray[Float64], sampling_rate: int, cutoff: flo
         :param ripple: ripple value for Chebyshev filter
         :type ripple: float
         """
+        check_memory_layout_row_major(data)
```
xloem, line 387: note that this is the same conceptual kind of check as https://github.com/brainflow-dev/brainflow/pull/253/files#diff-d5c3bcf84e585e262092315dad35c121847283e2084c58a0281297dde4cfcbf0R392 and applies to all the same places, so the code could be merged

tests/python/hardware_test.py
Andrey1994: yep, its intentional

python-package/brainflow/utils.py
```diff
@@ -0,0 +1,10 @@
+import numpy as np
+
+from brainflow.board_shim import BrainFlowError
+from brainflow.exit_codes import BrainflowExitCodes
+
+
+def check_memory_layout_row_major(data):
+    if not data.flags['C_CONTIGUOUS']:
+        raise BrainFlowError('wrong memory layout, should be row major, make sure you didnt tranpose array',
+                             BrainflowExitCodes.INVALID_ARGUMENTS_ERROR.value)
```
xloem, line 10: so personally I would combine them something like:
```
def check_memory_layout_row_major(data, ndim):
    if not data.flags['C_CONTIGUOUS']:
        raise BrainFlowError('wrong memory layout, should be row major, make sure you didn't transpose array',
                             BrainflowExitCodes.INVALID_ARGUMENTS_ERROR.value)
    if len(data.shape) != ndim:
        raise BrainFlowError('wrong shape for data, should be {}d array'.format(ndim),
                                 BrainflowExitCodes.INVALID_ARGUMENTS_ERROR.value)
```

xloem (approved):
It's always great to add more checks, assuming deleting that file is correct.

I have a couple medical appointments today and may not address other queued things for a bit.

PR #261, opened 2021-03-13
Julia package: make sure that the license file is in the same directory as the Project.toml file

DilumAluthge:
cc: @Andrey1994


Andrey1994:
thanks!

PR #270, opened 2021-04-14
adding gforce dual

Andrey1994:

### DIFF
--- a/src/board_controller/brainflow_boards.cpp
+++ b/src/board_controller/brainflow_boards.cpp
@@ -254,7 +254,7 @@ json brainflow_boards_json = {
         }},
         {"16",
             {{"name", "GforcePro"},
-            {"sampling_rate", 500},
+            {"sampling_rate", 650},
             {"package_num_channel", 0},
             {"timestamp_channel", 9},
             {"marker_channel", 10},
@@ -285,12 +285,12 @@ json brainflow_boards_json = {
         }},
         {"19",
             {{"name", "GforceDual"},
-            {"sampling_rate", 500}, // todo check sampling rate
+            {"sampling_rate", 650}, // todo check sampling rate
             {"timestamp_channel", 3},
             {"marker_channel", 4},
             {"package_num_channel", 0},
             {"num_rows", 5},
-            {"eeg_channels", {1, 2}}
+            {"emg_channels", {1, 2}}
         }}
     }
 }};
--- a/src/board_controller/brainflow_boards.cpp
+++ b/src/board_controller/brainflow_boards.cpp
@@ -285,7 +285,7 @@ json brainflow_boards_json = {
         }},
         {"19",
             {{"name", "GforceDual"},
-            {"sampling_rate", 650}, // todo check sampling rate
+            {"sampling_rate", 500},
             {"timestamp_channel", 3},
             {"marker_channel", 4},
             {"package_num_channel", 0},
--- a/src/board_controller/brainflow_boards.cpp
+++ b/src/board_controller/brainflow_boards.cpp
@@ -282,6 +282,15 @@ json brainflow_boards_json = {
             {"num_rows", 8},
             {"eeg_channels", {1, 2, 3, 4}},
             {"eeg_names", "T3,T4,O1,O2"}
+        }},
+        {"19",
+            {{"name", "GforceDual"},
+            {"sampling_rate", 500}, // todo check sampling rate
+            {"timestamp_channel", 3},
+            {"marker_channel", 4},
+            {"package_num_channel", 0},
+            {"num_rows", 5},
+            {"eeg_channels", {1, 2}}
         }}
     }
 }};
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,6 +233,7 @@ set (BOARD_CONTROLLER_SRC
     ${CMAKE_HOME_DIRECTORY}/src/board_controller/mit/fascia.cpp
     ${CMAKE_HOME_DIRECTORY}/src/board_controller/neurosity/notion_osc.cpp
     ${CMAKE_HOME_DIRECTORY}/src/board_controller/oymotion/gforce_pro.cpp
+    ${CMAKE_HOME_DIRECTORY}/src/board_controller/oymotion/gforce_dual.cpp
     ${CMAKE_HOME_DIRECTORY}/src/board_controller/ironbci/ironbci.cpp
     ${CMAKE_HOME_DIRECTORY}/src/board_controller/freeeeg32/freeeeg32.cpp
     ${CMAKE_HOME_DIRECTORY}/src/board_controller/neuromd/brainbit_bled.cpp
--- a/csharp-package/brainflow/brainflow/board_controller_library.cs
+++ b/csharp-package/brainflow/brainflow/board_controller_library.cs
@@ -64,7 +64,9 @@ namespace brainflow
         NOTION_2_BOARD = 14,
         IRONBCI_BOARD = 15,
         GFORCE_PRO_BOARD = 16,
-        FREEEEG32_BOARD = 17
+        FREEEEG32_BOARD = 17,
+        BRAINBIT_BLED_BOARD = 18,
+        GFORCE_DUAL_BOARD = 19
     };
 
 
--- a/java-package/brainflow/src/main/java/brainflow/BoardIds.java
+++ b/java-package/brainflow/src/main/java/brainflow/BoardIds.java
@@ -28,7 +28,9 @@ public enum BoardIds
     NOTION_2_BOARD (14),
     IRONBCI_BOARD (15),
     GFORCE_PRO_BOARD (16),
-    FREEEEG32_BOARD (17);
+    FREEEEG32_BOARD (17),
+    BRAINBIT_BLED_BOARD (18),
+    GFORCE_DUAL_BOARD (19);
 
     private final int board_id;
     private static final Map<Integer, BoardIds> bi_map = new HashMap<Integer, BoardIds> ();
--- a/julia-package/brainflow/src/board_shim.jl
+++ b/julia-package/brainflow/src/board_shim.jl
@@ -24,6 +24,8 @@ export BrainFlowInputParams
     IRONBCI_BOARD = 15
     GFORCE_PRO_BOARD = 16
     FREEEEG32_BOARD = 17
+    BRAINBIT_BLED_BOARD = 18
+    GFORCE_DUAL_BOARD = 19
 
 end
 
--- a/matlab-package/brainflow/BoardIDs.m
+++ b/matlab-package/brainflow/BoardIDs.m
@@ -22,5 +22,7 @@ classdef BoardIDs < int32
         IRONBCI_BOARD(15)
         GFORCE_PRO_BOARD(16)
         FREEEEG32_BOARD(17)
+        BRAINBIT_BLED_BOARD(18)
+        GFORCE_DUAL_BOARD(19)
     end
 end
--- a/python-package/brainflow/board_shim.py
+++ b/python-package/brainflow/board_shim.py
@@ -40,6 +40,8 @@ class BoardIds(enum.IntEnum):
     IRONBCI_BOARD = 15  #:
     GFORCE_PRO_BOARD = 16  #:
     FREEEEG32_BOARD = 17  #:
+    BRAINBIT_BLED_BOARD = 18  #:
+    GFORCE_DUAL_BOARD = 19  #:
 
 
 class LogLevels(enum.IntEnum):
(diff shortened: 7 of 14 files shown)


Andrey1994:
to build with oymotion devices you need to use this file https://github.com/brainflow-dev/brainflow/blob/master/tools/build_win64_vs19_oymotion.cmd

matthijscox:
Cool, as discussed I should put some print statements around this location (assuming the start_stream doesn't fail earlier).
https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L175

I was thinking about something like this to see the binary data: 
`for(size_t i=0;i<data->size();i++) printf("%x, ", (int)ptr[i]);`

I was also wondering if there is a way to compile faster (only these oymotion drivers), so I can iterate faster (I'm going to fail a lot...). Compilation took 1-2 minutes now I had the feeling, so I can only run max 15 experiments per 30 minutes. Which is often my only daily time duration for this ;)

I'll just dump my findings and thoughts in this PR if that's OK

matthijscox:
Btw I also have the GForce Pro+, which also failed to stream data, I might start with that one...

Andrey1994:
In cmd file which you use for compilation there is a command like rmdir build64, you can remove this one and it will be compiled faster

Andrey1994:
What is gforcepro+?

matthijscox:
> What is gforcepro+?

It's the new version of the gForce Pro. I tried to connect it using the GFORCE_PRO_BOARD, but it failed upon start_stream.

Andrey1994:
Interesting, for me it works fine for old gforcepro and I thought there are no software changes

Andrey1994:
I checked md5sum for gforceSDK dlls I use in brainflow and in their repo(latest master https://github.com/oymotion/gForceSDKCXX) and its the same, so we are using the latest version

matthijscox:
Ok, I will try to figure out what goes wrong then with this new gForce Pro+ armband, maybe the debug messages tell me something already. The device streamed data on their smartphone app, so it should all work fine.

I recall I simply didn't receive any data in brainflow.start_stream for 10 seconds or so and it then closed automatically.

Andrey1994:
https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L16 uncomment this line and add prints to callback functions to check that they are triggered. Right here https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L16

also in feature map you can enable smth else to confirm that there are callbacks for other datatypes https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L222

matthijscox:
Ok I enabled the gForce logging and added a logging->info to the onExtendedDeviceData callback.

I inserted the OYMotion dongle and turned on the gForce Pro+.

I prepare the session and get this info:
```
using BrainFlow
BrainFlow.enable_dev_logger(BrainFlow.BOARD_CONTROLLER)
board_shim = BrainFlow.BoardShim(BrainFlow.GFORCE_PRO_BOARD, params)
BrainFlow.prepare_session(board_shim)
[2021-04-20 11:34:38.556] [GForceHandleLogger] [info] onDeviceConnected
[2021-04-20 11:34:38.559] [GForceHandleLogger] [info] set FeatureMap
[2021-04-20 11:34:38.584] [GForceHandleLogger] [info] onOrientationData
...
[2021-04-20 11:34:38.755] [GForceHandleLogger] [info] onOrientationData
[2021-04-20 11:34:38.914] [GForceHandleLogger] [info] setDataNotifSwitch result: sucess
[2021-04-20 11:34:38.914] [GForceHandleLogger] [info] onExtendedDeviceData
[2021-04-20 11:34:38.915] [GForceHandleLogger] [trace] data received but acqusition is off.
... 
[2021-04-20 11:34:39.510] [GForceHandleLogger] [info] onExtendedDeviceData
[2021-04-20 11:34:39.510] [GForceHandleLogger] [trace] data received but acqusition is off.
[2021-04-20 11:34:43.653] [GForceHandleLogger] [info] setEMGRawDataConfig result: failed
... 
[2021-04-20 11:35:04.979] [GForceHandleLogger] [info] onExtendedDeviceData
[2021-04-20 11:35:04.979] [GForceHandleLogger] [trace] data received but acqusition is off.
```

This doesn't sound good? **setEMGRawDataConfig result: failed**

Then upon upon start_stream I get no data:
```
julia> BrainFlow.start_stream(board_shim)
[2021-04-20 11:35:22.996] [board_logger] [debug] use stub streamer
[2021-04-20 11:35:22.996] [GForceHandleLogger] [info] starting acqusition.
[2021-04-20 11:35:28.009] [board_logger] [error] no data received in 5 sec, stopping thread
[2021-04-20 11:35:28.032] [GForceHandleLogger] [info] stop acqusition.
ERROR: BrainFlow.BrainFlowError("Error in start_stream 18", 18)
```

matthijscox:
It's funny, but it seems the dual channel sensor actually just works. At least I can start a stream and get board data.

Andrey1994:
```
[2021-04-20 11:34:39.510] [GForceHandleLogger] [info] onExtendedDeviceData
[2021-04-20 11:34:39.510] [GForceHandleLogger] [trace] data received but acqusition is off.
```

It means that prepare_session was called but start_stream was not, and in fact its fine because there is a small delay between these 2 calls

Flag is set inside this function:
```
int start_stream (void *param)
{
    if (!bInitialized)
    {
        return (int)BrainFlowExitCodes::BOARD_NOT_CREATED_ERROR;
    }
    gforceHandle->logger->info ("starting acqusition.");
    bShouldStopStream = false;

    return (int)BrainFlowExitCodes::STATUS_OK;
}
```

I think somewwhere in log messages you should see this one 

```
logger->trace ("EMG is not configured.");
```

<< This doesn't sound good? setEMGRawDataConfig result: failed

Yes, it seems like the main issue. Also, I've just realized that EMG Configuration more likely should be different for pro and dual board...

I think here:

```
ds->setEMGRawDataConfig (GforceHandle::iSamplingRate, // sample rate
            (DeviceSetting::EMGRowDataChannels) (0x00FF),     // channel 0~7
            GforceHandle::iTransactionSize,                   // data length
            GforceHandle::iADCResolution,                     // adc resolution
            [this] (ResponseResult result) {
                std::string res =
                    (result == ResponseResult::RREST_SUCCESS) ? ("sucess") : ("failed");
                bIsEMGConfigured = (result == ResponseResult::RREST_SUCCESS) ? (true) : (false);
                this->logger->info ("setEMGRawDataConfig result: {}", res);
            });
    }
```
must be if statement for different devices.

I guess for dual board instead `0x00FF` it should be smth different, sampling rate and transaction size can also be different... 
But since dual board just works for you probably we can keep as is

For pro+ issue we can:

* try to find more info about emg configuration on their docs
* ask oymotion guys
* ignore check like `if(emgconfigured)` and see what will happen

matthijscox:
Hmm Dual board is not stable with this configuration. Now it doesnt start streaming anymore.
Upon prepare_session I also get this error/warning:
`(2021-4-20 13:06:35:951)[E/gForceSDK]: Unknown data type: 0x7F`

Restarted everything:
* Closed Julia REPL
* Turned off Dual sensor device
* Unmounted dongle
* Remounted dongle
* Turn on Dual sensor device
* Started Julia REPL

Now it streams again and I can `get_board_data` again.

matthijscox:
Ok, gForcePro+ seems to be fixed with these settings I found here:
https://oymotion.github.io/gForceSDK/gForceSDK/

```
    static const int iADCResolution = 8;
    static const int iTransactionSize = 128;
    static const int iSamplingRate = 650;
```

Data looks sensible in my little Julia visualizer:

image

Andrey1994:
So, new gforcepro+ doesnt work with 12 bits resolution?

Andrey1994:
For 8bits data should be parse differently, 

instead 
```
emgData[i + 1] = (double)*(reinterpret_cast<const uint16_t *> (ptr));
ptr += 2;
```
it should be smth like
```
emgData[i + 1] = (double)*(reinterpret_cast<const uint8_t *> (ptr++));
```

And probably size will be different as well. I will test it with old gforcepro and update PR

src/board_controller/brainflow_boards.cpp
```diff
@@ -282,6 +282,15 @@ json brainflow_boards_json = {
...
             {"eeg_names", "T3,T4,O1,O2"}
+        }},
+        {"19",
+            {{"name", "GforceDual"},
+            {"sampling_rate", 500}, // todo check sampling rate
+            {"timestamp_channel", 3},
+            {"marker_channel", 4},
+            {"package_num_channel", 0},
+            {"num_rows", 5},
+            {"eeg_channels", {1, 2}}
```
matthijscox, line 293: emg_channels

Andrey1994:
julia> BrainFlow.get_current_board_data(512, board_shim)
11×48 transpose(::Matrix{Float64}) with eltype Float64:

julia> BrainFlow.get_current_board_data(12, board_shim)
11×12 transpose(::Matrix{Float64}) with eltype Float64:

it should not be 11 here for gforce dual, it must be 5. Are you sure you are using correct board id?

matthijscox:
Ok, I now did the following as an experiment:

Changed the channel mask to 2:
```
        ds->setEMGRawDataConfig (GforceHandle::iSamplingRate, // sample rate
            (DeviceSetting::EMGRowDataChannels) (0x0002),
```     

Added some printing of the package data:
```
        if (iBoardType == (int)BoardIds::GFORCE_DUAL_BOARD)
        {
            printf("Package data:\n[\n");
            for(size_t i=0;i<data->size();i++) printf("%x\n", (int)ptr[i]);
            printf("]\n");
```

Then I get the following data in one package:
```
[
fe
0
0
0
0
11
f9
fd
fd
fd
fd
fd
fe
fe
0
0
0
0
11
f9
fd
fd
fd
fe
fd
fe
fe
0
0
0
0
11
f3
fd
fd
fd
fd
fe
fd
fe
0
0
0
0
11
f2
fd
fd
fd
fd
fd
fe
fe
0
0
0
0
10
f3
fd
fd
fd
fd
fd
fe
fe
0
0
0
0
10
f5
fd
fd
fd
fd
fd
fd
fe
0
0
0
0
f
fe
fd
fd
fd
fe
fd
fd
fe
0
0
0
0
e
fe
fd
fd
fd
fd
fd
fe
fe
0
0
0
0
d
fe
fd
fd
fd
fd
fd
fd
fe
0
0
0
0
b
fe
fd
fd
fd
fd
]
```

Also note I do get 5 channels (I was using the wrong board_id in the past experiments, please ignore):
```
julia> data = BrainFlow.get_board_data(board_shim)
5×1344 transpose(::Matrix{Float64}) with eltype Float64:
````

Andrey1994:
I added 8 bits resolution and also fixed channel map. It should be 0x0002 | 0x0001 for dual board

matthijscox:
I got this reply from Hualiang for gForceDual 
`setEmgDataConfig(500, 0x03, 32, 8, callback);`

Sampling rate is configurable, so maybe we can keep 500, or we'd like 650?
I guess package size of 32 is also configurable.

Andrey1994:
I changed params for gforcedual in latest commit(I hope I didnt make a mistake)

What I dont like about these changes is 8 bits resolution, I think 8 bits instead 12 lead to worse accuracy of classifiers...
Would be good to ask oymotion guys why 12 bits doesnt work for pro+

matthijscox:
gForceDual seems to work now, streaming data (without connecting it to myself):
image

matthijscox:
Hualiang says the gForcePro+ should support 12bit ADC if limited to 500Hz sampling rate. It limited to BLE bandwidth. But that's the settings we used in the past.

I could try to play with even lower sampling rate, to see if the setEMGRawDataConfig doesn't fail.

Andrey1994:
<< gForceDual seems to work now

Great! Lets test a few more times and with real data and create new release

For gforcepro+ lets try 500 and 12 one more time. Sampling rates lower than 500 are not good for emg. Also, we can try 500 and 12 for gforcedual

matthijscox:
I tried 500Hz and 12bit.
gForceDual seemed to connect fine, though lot's of trace messages.
gForcePro+ connected once, but got "setEMGRawDataConfig result: failed" on 3 out of 4 attempts.

The OYMotion CTO emailed me and said the command may fail if the data transfer is in progress, so he adviced to call setDataNotifSwitch() in the callback of setEMGRawDataConfig(). He might join this PR and explain and/or make a commit.

Andrey1994:
<<He might join this PR and explain and/or make a commit.
it would be good since I dont fully understand idea about callback

we send command for 8 bits and they are fine all time and for 12 and they fail randomly.. I think data transfer should affect both options the same way

PR #271, opened 2021-04-16
Revert "refactor cyton status check, add log message"

Andrey1994:
Reverts brainflow-dev/brainflow#269

In code #269 looks much better but it works worse, revert


Andrey1994:
https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L16 uncomment this line and add prints to callback functions to check that they are triggered. Right here https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L16 

also in feature map you can enable smth else to confirm that there are callbacks for other datatypes https://github.com/brainflow-dev/brainflow/blob/master/third_party/gForceSDKCXX/src/inc/gforce_handle.h#L222

PR #282, opened 2021-05-24
change oymotion resolution to 12bits

Andrey1994:

Restore 12bits resolution for oymotion devices(based on email thread). @matthijscox could you review and check that it works fine?

### DIFF
--- a/python-package/examples/plot_real_time.py
+++ b/python-package/examples/plot_real_time.py
@@ -20,7 +20,7 @@ class Graph:
         self.board_shim = board_shim
         self.exg_channels = BoardShim.get_exg_channels(self.board_id)
         self.sampling_rate = BoardShim.get_sampling_rate(self.board_id)
-        self.update_speed_ms = 10
+        self.update_speed_ms = 50
         self.window_size = 4
         self.num_points = self.window_size * self.sampling_rate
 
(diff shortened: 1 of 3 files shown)


matthijscox:
I used this branch to compile the libs. 

Successfully connected the GFORCE_PRO_BOARD, could get board data without any sign of problems.

Then stopped/released/disconnected everything, including the dongle. 
Reconnected with the GFORCE_DUAL_BOARD. Could again succesfully connect and get board data.

@Andrey1994 it looks good from my side!

matthijscox (approved):
All good

Andrey1994:
Great! To double check are you sure that correct libraries were used? Emg values should be greater than 256

Andrey1994:
And did you test it using new gforce pro?

matthijscox:
Yes, this is the gForcePro+, which works fine with the GFORCE_PRO_BOARD.

Here's a snippet of data, but that's already in microvolts right? Can you already see from this if it surpasses the 8bits limit?
image

Just to try, I did a unique, and there are more than 256 types of values:
image

Andrey1994:
<<but that's already in microvolts right? Can you already see from this if it surpasses the 8bits limit?

its definitely 12 bits(for 8 bits max value here is 256, or maybe even 128 if its signed) .
And its the only board where we return non uV value, mostly because I dont know equation to convert it to uV. I asked about it here https://github.com/oymotion/gForceSDKCXX/issues/4 but still dont fully understand how to convert that

Andrey1994:
does it also work for gforce dual?

Andrey1994:
And when you run `import BrainFlow` you saw "using local libraries" didnt you?

matthijscox:
Yes it works for gForceDual and yes I saw the local library usage. I copied the local /lib from the build process to the julia brainflow folder.

PR #301, opened 2021-06-28
CMake refactoring

Andrey1994:


matthijscox:
Works for me on Windows.
Is there still a way to speed it up, like with the bash scripts where I could edit to not remove the build64 folder?

image

matthijscox:
> Is there still a way to speed it up, like with the bash scripts where I could edit to not remove the build64 folder?

Wait, I think I see the --clear-build-dir flag. If not provided, it'll re-use the existing build dir I assume.

Andrey1994:
yes, bu default it will not clean build folder and will fork faster

PR #307, opened 2021-07-09
🚀 Neurosity Crown Support

alexcastillo:


src/utils/inc/brainflow_constants.h
```diff
@@ -56,6 +56,7 @@ enum class BoardIds : int
     GALEA_SERIAL_BOARD = 20,
     MUSE_S_BLED_BOARD = 21,
     MUSE_2_BLED_BOARD = 22,
+    CROWN_BOARD = 23,
     // use it to iterate
     FIRST = PLAYBACK_FILE_BOARD,
     LAST = MUSE_2_BLED_BOARD
```
Andrey1994, line 62: need to update LAST here to point to CROWN_BOARD(its smth like a way to iterate via enum in c++)
Also, this enum is duplicated in bindings(enums cannot be passed from cpp). Need to add here https://github.com/brainflow-dev/brainflow/blob/master/python-package/brainflow/board_shim.py#L47 and in other places

alexcastillo: Done. Also added the board in other places where I found Notion boards.

PR #310, opened 2021-07-13
add gyro data for muse devices

Andrey1994:


Andrey1994:
@matthijscox could you check that it works for muse 2?

PR #319, opened 2021-08-10
Enable optical connection for FreeEEG32 on Windows

mesca:
The FreeEEG32 board supports both USB and optical connections. Unfortunately, the optical connection does not currently work on Windows (it works as expected on macOS). This is because the serial port settings are disabled for Windows. While a comment in the code indicates that the driver ignores those settings, I have found out that it is actually possible to set the baudrate to 921600 (at least on my system, using the [Silabs USB to UART driver](https://www.silabs.com/developers/usb-to-uart-bridge-vcp-drivers), which is required for the optical dongle).

### DIFF
--- a/src/board_controller/freeeeg32/freeeeg32.cpp
+++ b/src/board_controller/freeeeg32/freeeeg32.cpp
@@ -202,30 +202,19 @@ int FreeEEG32::open_port ()
 
 int FreeEEG32::set_port_settings ()
 {
-#ifdef _WIN32
-    // windows driver fails to set settings and in fact ignores them, no idea what drivers on others
-    // OSes do
-    int timeout_only = true;
-#else
-    int timeout_only = false;
-#endif
-    int res = serial->set_serial_port_settings (1000, timeout_only);
+    int res = serial->set_serial_port_settings (1000, false);
     if (res < 0)
     {
         safe_logger (spdlog::level::err, "Unable to set port settings, res is {}", res);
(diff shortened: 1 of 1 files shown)


Andrey1994:
Thanks!

I have no freeeeg32 anymore and cannot test... But if I remember correctly I saw some errors on windows. Maybe I used another driver(there is a lack of documentation for freeeeg32 and no idea which one is recommended)

Could you change it to call these methods for windows but don't throw an exception(don't set error code) for windows?
Smth like: 
```
#ifndef _WIN32
return (int)BrainFlowExitCodes::SET_PORT_ERROR;
#endif
```

mesca:
Done.

src/board_controller/freeeeg32/freeeeg32.cpp
```diff
@@ -202,30 +202,24 @@ int FreeEEG32::open_port ()
...
-    int timeout_only = true;
-#else
-    int timeout_only = false;
-#endif
-    int res = serial->set_serial_port_settings (1000, timeout_only);
+    int res = serial->set_serial_port_settings (1000, false);
     if (res < 0)
     {
         safe_logger (spdlog::level::err, "Unable to set port settings, res is {}", res);
         return (int)BrainFlowExitCodes::SET_PORT_ERROR;
```
Andrey1994, line 209: I think this one should be under `#ifndef _WIN32` too

mesca: Fixed.

PR #325, opened 2021-08-14
added get_board_data(int num_datapoints) method 

RAMESSESII2:
Hey, @Andrey1994 I've tried adding what I believe you're saying. Do guide me if there's anything wrong with it. Thanks.
As mentioned in the issue, I've added an overloaded method to allow users to specify the number of datapoints.


cpp-package/src/inc/board_shim.h
```diff
@@ -201,6 +201,7 @@ class BoardShim
     int get_board_data_count ();
     /// get all collected data and flush it from internal buffer
     BrainFlowArray<double, 2> get_board_data ();
+    BrainFlowArray<double, 2> get_board_data (int);
```
Andrey1994, line 204: need to add comment for doxygen like above and it will be displayed in docs properly

cpp-package/src/board_shim.cpp
```diff
@@ -168,6 +168,22 @@ BrainFlowArray<double, 2> BoardShim::get_board_data ()
     delete[] buf;
     return matrix;
 }
+BrainFlowArray<double, 2> BoardShim::get_board_data (int num_samples)
+{
```
Andrey1994, line 172: need to add a check that num_samples is less or equal to the current size of a buffer(check original implementation of get_board_data method)

Andrey1994: if num_samples is bigger need to return std::min(num_samples, current_size)

Andrey1994 (commented):
Hi and thanks for starting it!
In addition to comments in code I also need to mention that it should be done for each programming language supported(python,java,etc) Link for python https://github.com/brainflow-dev/brainflow/blob/master/python-package/brainflow/board_shim.py#L1034

cpp-package/src/inc/board_shim.h
RAMESSESII2: sure

cpp-package/src/board_shim.cpp
RAMESSESII2: Done, hopefully, in the latest commit, I've addressed them. Also, I've added the overloaded function for Java, C++, and Python. Please check and suggest if these are good to go. And then I can work upon other languages.

cpp-package/src/board_shim.cpp
```diff
@@ -168,6 +168,23 @@ BrainFlowArray<double, 2> BoardShim::get_board_data ()
     delete[] buf;
     return matrix;
 }
+BrainFlowArray<double, 2> BoardShim::get_board_data (int num_datapoints)
+{
+    int num_samples = get_board_data_count ();
+    num_samples = std::min(num_samples, num_datapoints);
```
Andrey1994, line 174: there is a CI pipeline that checks code style and it will be failed, also there is a tool to format code automatically, take a look at https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style

Andrey1994: and let's add a check that this value is bigger than 0(the same for all languages), otherwise, it should throw an exception with INVALID_ARGUMENTS_ERROR code

java-package/brainflow/src/main/java/brainflow/BoardShim.java
```diff
@@ -793,4 +793,23 @@ public boolean is_prepared () throws BrainFlowError
         }
         return result;
     }
+
+    public double[][] get_board_data (int num_datapoints) throws BrainFlowError
+    {
+        int size = get_board_data_count ();
+        size = (size >= num_datapoints)? num_datapoints: size;
```
Andrey1994, line 800: add check for positive value here and for other languages

Andrey1994 (commented):
Looks good so far, thanks

Let's also use this new method in one of the tests, for example in downsampling https://github.com/brainflow-dev/brainflow/blob/master/tests/cpp/signal_processing_demo/src/downsampling.cpp(there are similar tests for other languages)

Andrey1994:
there is a failed test for python because there overloading should be done using default value

RAMESSESII2:
> Looks good so far, thanks
> 
> Let's also use this new method in one of the tests, for example in downsampling https://github.com/brainflow-dev/brainflow/blob/master/tests/cpp/signal_processing_demo/src/downsampling.cpp (there are similar tests for other languages)

I'm not sure how to do that. 
Though I've added checks for the Overloaded functions to raise an exception if datapoints passed are not greater than 0 and have resolved the problem with the python overloaded method by passing a positional argument.

Andrey1994:
"
I'm not sure how to do that.
Though I've added checks for the Overloaded functions to raise an exception if datapoints passed are not greater than 0 and have resolved the problem with the python overloaded method by passing a positional argument.
"

This samples calls method `get_current_board_data (num)`  you can replace it by `get_board_data(num)`

RAMESSESII2:
> This sample calls method `get_current_board_data (num)` you can replace it by `get_board_data(num)`

RAMESSESII2:
> This sample calls method `get_current_board_data (num)` you can replace it by `get_board_data(num)`

Ok, sure I can do that.

Andrey1994:
"Also, can you guide me towards how to see "CHECKS" passed locally? I'm using Fedora 33."

its a little more complicated to setup, the easiest option is to push your changes and see failed tests and after that replicate the exact test locally

Andrey1994:
All tests and checks in CI can be found here https://github.com/brainflow-dev/brainflow/tree/master/.github/workflows

Andrey1994:
Looks good so far, lets add the remaining languages and a few tests

RAMESSESII2:
I'm on it.

cpp-package/src/board_shim.cpp
```diff
@@ -168,6 +168,29 @@ BrainFlowArray<double, 2> BoardShim::get_board_data ()
     delete[] buf;
     return matrix;
 }
+BrainFlowArray<double, 2> BoardShim::get_board_data (int num_datapoints)
+{
+
+    if (num_datapoints <= 0)
+    {
+        throw std::invalid_argument (
```
Andrey1994, line 176: change that to BrainFlowException and the second argument should be `(int)BrainFlowExitCodes::INVALID_ARGUMENTS_ERROR`

java-package/brainflow/src/main/java/brainflow/BoardShim.java
```diff
@@ -793,4 +793,26 @@ public boolean is_prepared () throws BrainFlowError
         }
         return result;
     }
+
+    public double[][] get_board_data (int num_datapoints) throws BrainFlowError
+    {
+        if(num_datapoints <= 0){
+            throw new IllegalArgumentException ("data size should be greater than 0");
```
Andrey1994, line 800: the same here

julia-package/brainflow/src/board_shim.jl
```diff
@@ -229,8 +229,13 @@ end
     return sub_string
 end
 
-@brainflow_rethrow function get_board_data(board_shim::BoardShim)
+@brainflow_rethrow function get_board_data(board_shim::BoardShim, num_samples::Int64=1)
     data_size = get_board_data_count(board_shim)
+    if num_samples != nothing
+        if num_samples <= 0
+            throw(DomainError(num_samples, " data size should be greater than 0."))
```
Andrey1994, line 236: here and in all other bindings

julia-package/brainflow/src/board_shim.jl
```diff
@@ -229,8 +229,13 @@ end
     return sub_string
 end
 
-@brainflow_rethrow function get_board_data(board_shim::BoardShim)
+@brainflow_rethrow function get_board_data(board_shim::BoardShim, num_samples::Int64=1)
```
Andrey1994, line 232: 1 is bad default because it changes the behavior of existing code. Currently users have `get_board_data()` and it returns all available data, such behavior should not be changed. Set default to -1 for example or to 0 and handle it to return all data.

Or add new methods if its allowed in Julia(I don't remember, you can check or ask @matthijscox )

matlab-package/brainflow/BoardShim.m
```diff
@@ -377,10 +377,15 @@ function insert_marker(obj, value)
             num_data_point = data_count.value;
         end
 
-        function data_buf = get_board_data(obj)
-            % get all collected data and remove it from the buffer
+        function data_buf = get_board_data(obj, num_data_point)
```
Andrey1994, line 380: I don't get where  a default value is

python-package/brainflow/board_shim.py
```diff
@@ -46,6 +47,7 @@ class BoardIds(enum.IntEnum):
     MUSE_S_BLED_BOARD = 21  #:
     MUSE_2_BLED_BOARD = 22  #:
     CROWN_BOARD = 23  #:
+
```
Andrey1994, line 50: new line here is not needed, it seems like in python you have some autoformatter, I am fine with it but here new line looks bizarre

Andrey1994 (commented):
looks good, need to change the exception type, a few defaults, and update one test to use this new method

RAMESSESII2:
hey thanks a lot, 've been busy with my college and studies. I can see you've already corrected most of my blunders. 👍🏼

Andrey1994:
Yes, I helped you a little with it. Thanks for implementing it!

PR #334, opened 2021-08-25
Added Live Streaming Example for Enophone

ddoyon92:
Live streaming example with basic processing and signal quality feedback.


third_party/enophone/enophone_streaming.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class Graph:
+    def __init__(self, board_shim):
+        self.board_id = board_shim.get_board_id()
+        self.board_shim = board_shim
+        self.exg_channels = BoardShim.get_exg_channels(self.board_id)
+        self.sampling_rate = BoardShim.get_sampling_rate(self.board_id)
+        self.update_speed_ms = 50
+        self.window_size = 2500
```
Andrey1994, line 23: for convenience lets use widnow size in seconds and add smth like `num_points = window_size * BoardShim.get_sampling_rate(BoardIds.ENOPHONE_BOARD)`

third_party/enophone/enophone_streaming.py
```diff
@@ -0,0 +1,129 @@
...
+                    self.legends[count].setLabelTextColor('r')
+
+        self.app.processEvents()
+
+
+def main():
+    BoardShim.enable_dev_board_logger()
+    logging.basicConfig(level=logging.DEBUG)
+
+    parser = argparse.ArgumentParser()
```
Andrey1994, line 85: you can remove all arguments and keep only mac address

Andrey1994 (commented):
third_party folder is for c++ dependencies used inside brainflow and not written by me. Let's move it to python-package/examples

third_party/enophone/enophone_streaming.py
ddoyon92: Done.

PR #347, opened 2021-09-17
Improve const correctness, remove const_cast

serense:


Andrey1994:
Hi and thanks for this fix.

In methods exposed from dlls it was `char *` instead `const char *` because I was not sure that all languages support calling methods with `const char *`(especially smth like matlab which parses headers using regexps). Tests are passed and it's great.

Andrey1994 (approved):

PR #349, opened 2021-09-23
fixed most of security warnings

khoinguyen-hub:
@Andrey1994  I've fixed most of the warnings. The ones that have to do with exception; however, I don't really know how to handle them so I put them as `except BaseException:`. I would love to hear any feedback and any tips or suggestions to change the code before the merge.

### DIFF
--- a/src/board_controller/inc/brainflow_input_params.h
+++ b/src/board_controller/inc/brainflow_input_params.h
@@ -30,7 +30,7 @@ struct BrainFlowInputParams
     }
 
     // default copy constructor and assignment operator are ok, need less operator to use in map
-    bool operator< (const struct BrainFlowInputParams &other) const
+    bool operator< (struct BrainFlowInputParams other) const
     {
         return std::tie (serial_port, mac_address, ip_address, ip_port, ip_protocol, other_info,
                    timeout, serial_number, file) <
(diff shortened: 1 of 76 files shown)


Andrey1994:
Thanks, I will review it tomorrow, in the meantime you can fix failed tests. For clang-format take a look at https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style

Andrey1994:
Also, CI compiles it with warnings as errors option, to reproduce it locally you can add `--warnings-as-errors` argument to build.py

khoinguyen-hub:
I'm having trouble with Cmake and building it in general. I think because I don't have Visual Studio 2019 installed which I am getting it now.

src/board_controller/inc/brainflow_input_params.h
```diff
@@ -30,25 +30,25 @@ struct BrainFlowInputParams
     }
 
     // default copy constructor and assignment operator are ok, need less operator to use in map
-    bool operator< (const struct BrainFlowInputParams &other) const
```
Andrey1994, line 33: why is it changed?

src/ml/inc/brainflow_model_params.h
```diff
@@ -21,7 +21,7 @@ struct BrainFlowModelParams
     }
 
     // default copy constructor and assignment operator are ok, need less operator to use in map
-    bool operator< (const struct BrainFlowModelParams &other) const
```
Andrey1994, line 24: here too

Andrey1994 (commented):
Looks good so far, let's fix the remaining recommendations

Andrey1994:
Exception handling is fine

src/ml/inc/brainflow_model_params.h
khoinguyen-hub: @Andrey1994 i think i changed these on accdient while changing the other functions arguments.

khoinguyen-hub: They should still be const and pass by value.

Andrey1994: Could you restore it here and in other places?

khoinguyen-hub: ok i will change them back

Andrey1994 (approved):

Andrey1994:
Thanks for fixing it, there are a few more recommendations from lgtm you are welcome to fix them in another PR

PR #356, opened 2021-10-07
Add stateful filters to data_handler interface

serense:
The goal is to provide better support for online filtering by saving filter state instead of using one shot perform_{lowpass, highpass, ...} functions.

### DIFF
--- a/.github/workflows/run_unix.yml
+++ b/.github/workflows/run_unix.yml
@@ -179,6 +179,8 @@ jobs:
         sudo -H python3 $GITHUB_WORKSPACE/tests/python/serialization.py
     - name: Filters Python
       run: sudo -H python3 $GITHUB_WORKSPACE/tests/python/signal_filtering.py
+    - name: Filters with state Python
+      run: sudo -H python3 $GITHUB_WORKSPACE/tests/python/signal_filtering_with_state.py
     - name: Transforms Python
       run: sudo -H python3 $GITHUB_WORKSPACE/tests/python/transforms.py
     - name: Downsampling Python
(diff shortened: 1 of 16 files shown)


Andrey1994:
I can help with some bindings if you want me to. BTW are you a member of slack workspace?

serense:
I've changed the API to use an `int` instead of a pointer to the filter object. Also added a basic python binding.

I understand that exposing a similar API in all languages is desirable. However I don't have a local development environment for any language except C/C++ and python which makes the task more or less infeasible. Also there doesn't seem to be an easy way to run all tests locally which makes the whole process really cumbersome.

Regarding the motivation for the added feature, the new functionality is more fundamental than the one already provided. I believe the current functionality requires more knowledge from the user. Windowing and filtering means you introduce transients at the start of each window. Working around those transients in a reasonable way requires knowledge about the filter response. There is also an overhead of the current functionality not only in needing to use overlapping windows to compensate for transients, but also because the filters are allocated on the stack which require repeated memory allocations. For me, the main motivation is the more straightforward use rather than the overhead.

serense:
Any help implementing bindings for different languages would be appreciated.

I did sign up for the Slack workspace but I haven't been active.

Andrey1994:
@trobanga could you do it for Rust? Also, feel free to update some other languages if you want to and\or review, it would be great

trobanga:
yes, I can do the Rust binding this week (probably). Other bindings I can review but I don't think I have the time and/or knowledge to write them.

trobanga:
If you use filters.size() as ID then filters can't be removed. Is that intended?

trobanga:
> If you use filters.size() as ID then filters can't be removed. Is that intended?

Ok, I see that now...however, why are you storing the filters in a global variable in the first place? Why not return the filter instead of the filter id? I think that would be more flexible.

PR #361, opened 2021-10-09
Rust Package: fix dynamic link loading

trobanga:
This PR fixes loading of dynamic libraries


Andrey1994:
I've merged #362 but didnt update rust binding there, could you change it in this PR as well? Or we can create a new one, its also ok

trobanga:
Ok, this should work now

PR #365, opened 2021-10-15
Rust CI and examples

Andrey1994:
@trobanga, when I am running test on Windows, the linking process is failed with `LINK : fatal error LNK1181: cannot open input file 'BoardController.lib' `

I have not investigated it but have a guess, on Windows for DLLs there are two files in fact .lib(which is a stub and used for linking) and .dll it seems like now rust binding uses linking instead of loading DLLs manually. Probably in build.cmake files need to copy .lib file also.

Also, there are some enums missed


Andrey1994:
Copying .lib files is not complicated but it also probably means that dll files should be inside the search path at runtime(the same for Unix I guess). So they should be copied to the same folder where exe is located or we need to set `PATH` or `LD_LIBRARY_PATH`

trobanga:
The libraries are copied into target/build/... so there shouldn't be any more action required from us.

bindgen is only needed to generate the bindings. This is only necessary if the API changes i.e. running `cargo build --features="generate_binding"` should only be done manually and one needs to adjust the source code afterwards. (It generates the files in src/ffi/).
So, I think it's fair to assume that any developer will either have clang installed or will be able to do so.

I've changed it such that it only adds the bindgen dependency if the generate_binging feature is selected.

Andrey1994:
Ah, files from ffi folder are also inside repo, so users dont need to run generate binding even for local installation, right?

Andrey1994:
We dont have static versions of these libs(of course we can add it) and do not have a global installer or smth that copies libs to a single location for all languages, we are trying to use only built-in language tools\build systems.

I am looking at https://users.rust-lang.org/t/linking-with-local-c-library/39866/4 and it seems like there are ways to tell rust about search paths for dyn libs

trobanga:
This is already done at the bottom of build.rs . And it definitely works using `cargo run` inside the project directory.
But calling the binary directly e.g. `./target/debug/bftest` then one needs to set LD_LIBRARY_PATH

Andrey1994:
@trobanga here is a list of problems I've found so far, I will update CI pipelines to run tests automatically and docs. And after that will continue looking at the issues below, would be good if you check as well 

* fft and wavelet transforms produce wrong results(res after inverse transform should be the same as before)
* downsampling test produce an empty array
* serialization test(write_file and read_file) fails to compile, its the reason I didn't add
* get_avg_bandpowers is wrong(should be used in test eeg_metrics), it should accept one more arg with channels, combine data from specified channels together into 1d vector and pass it to C\C++ code

PR #368, opened 2021-10-23
Rust ci

trobanga:
Change tests to actual Rust tests (can be used with `cargo test` now)

Fix downsampling and inverse wavelet transform


Andrey1994:
@trobanga thanks, please also fix CI pipeline for windows before merging

PR #376, opened 2021-11-01
Use ndarray::Array2 as data format

trobanga:
Add R/W file example


tests/rust/.gitignore
```diff
@@ -0,0 +1,20 @@
+# Created by https://www.toptal.com/developers/gitignore/api/rust
```
Andrey1994, line 1: maybe I am missing smth but in this folder, there is only .gitignore and cargo.toml, did you forgot to add smth?

rust-package/brainflow/examples/read-write-file.rs
```diff
@@ -0,0 +1,36 @@
...
+    let mut tmp_dir = env::temp_dir();
+    tmp_dir.push("brainflow_tests");
+    tmp_dir.push("rust");
+    fs::create_dir_all(&tmp_dir).unwrap();
+    tmp_dir.push("read-write_file.csv");
+
+    let filename = tmp_dir.to_str().unwrap();
+
+    dbg!(&data);
+    // let data = data
```
Andrey1994, line 29: rm comment

docs/UserAPI.rst
```diff
@@ -149,3 +149,8 @@ Example:
 
 .. literalinclude:: ../julia-package/brainflow/test/serialization.jl
    :language: julia
+
+Rust API Reference
```
Andrey1994, line 153: lets add docs as the latest step, in the current master its removed for now

.github/workflows/run_windows.yml
```diff
@@ -120,6 +120,11 @@ jobs:
       uses: julia-actions/setup-julia@v1
       with:
         version: 1.3.1
+    - name: Build Rust Test
```
Andrey1994, line 123: all examples are in another folder what does it test\build?

tests/rust/.gitignore
trobanga: I started with that branch yesterday and thought I could merge master into it...which was a bad idea. This folder should not exist...that's why I closed the PR. I'll make a new one

PR #377, opened 2021-11-01
Use ndarray as data format and add rw-file example

trobanga:


Andrey1994 (changes requested):
`get_current_board_data` method can return less datapoints than requested, its why there is argument to return len in low-level API. Need to use returned value in reshaping

PR #385, opened 2021-11-07
Use rust enums as function parameters

trobanga:
Add FromPrimitive and ToPrimitive to enum derives and use enums directly as function parameters


Andrey1994 (changes requested):
lets also change it here https://github.com/brainflow-dev/brainflow/blob/master/rust-package/brainflow/src/brainflow_input_params.rs#L12, it should be IpProtocolType enum

trobanga:
You are right, I've changed it for the Builder only because the parameters are serialized into a string anyway. Using it in the params directly requires some extra work because IpProtocolType does not derive all the necessary traits (yet)

Andrey1994:
also need to fix one merge conflict(just file renaming)

Andrey1994:
git still shows it as a conflict, you should merge master branch to your branch to solve it in git

Andrey1994:
I've disabled publishing to cargo for now(will fix later on) https://github.com/brainflow-dev/brainflow/commit/b646816da4d968118ff56d8e4a552aa36e4508de let's fix this enum and create an initial release for rust, other improvements can be made later after the official announcement

trobanga:
I would consider it fixed already because if one uses the builder pattern to construct the struct then the IpProtocol enum is used. 
I agree that it would be nicer if it would be used in the struct too but that would require an enormous amount of work compared with the gain.

Andrey1994:
Could you explain why it's more complicated here compared to BrainFlowModelParams? https://github.com/brainflow-dev/brainflow/blob/master/rust-package/brainflow/src/brainflow_model_params.rs it seems like these two cases are identical, its almost the same struct and it uses almost the same enums

trobanga:
BrainflowModelParams has less derive macros. For BrainflowModelParams I had only to implement Serialize manually (I skipped Deserialize for now). For BrainFlowInputParams I would have to do the same for Default, PartialEq, PartialOrd too or regex it into the output from build.rs. But then we'd need to import some functions too in ffi/constants.rs and so on...

I will write custom get/set functions for the ip_protocol field such that IpProtocolType is used there.

trobanga:
Ok, I will do it once the [getset](https://github.com/Hoverbear/getset) crate has its next release because I'm missing the #[getset(skip)] feature.

However, I think it's okay to publish it like this because the only way to construct BrainFlowInputParams is with the builder and there IpProtocolType is used.

Andrey1994:
Thanks!

PR #386, opened 2021-11-07
Info about rust release

Andrey1994:


trobanga:
The example needs an update according to #385:

`let board = board_shim::BoardShim::new(BoardIds::SyntheticBoard, params).unwrap();`

trobanga (approved):

Andrey1994:
publishing to crates.io is blocked by https://github.com/rust-lang/crates.io/issues/40 possible solutions are:

* consider other registries(https://doc.rust-lang.org/cargo/reference/registries.html), unfortunately, GitHub packages don't provide a registry for rust but maybe there are other options
* ask for a quote higher than 10MB 
* don't publish crates for now

trobanga:
> Building C\C++ code from Rust is not an option. 
Why not? In my first version I did exactly that.

trobanga:
- I don't think another registry is a good solution because crates.io is the default. I don't even know about others.
- it will be hard to argue for a higher quote because the preferred way is to build from source. I think it has many advantages over providing libs e.g. build for other platforms
- I also don't like the idea to dowload it from somewhere else

Andrey1994:
https://github.com/brainflow-dev/brainflow/pull/386#issuecomment-963839389 first of all because technically it's exactly the same as downloading binary files since source code should be downloaded from github first but in addition it requires compilation

Also, this way rust users will need to install CMake, Visual Studio, and some very specific components for Visual Studio(correct winsdk version, atl package, smth from UWP to work with BLE, etc). Finally, it's extra work to reproduce CMake parameters in Rust build script

trobanga:
But you are downloading ~30MB now of which 2 / 3 you don't need because they are for different platforms. If you do that for all your dependencies it will blow up.
Then the problem you are describing is specific to windows. One option could be to ship only Windows binaries and for other platforms build from source. I don't know how large the win bins are but with that it might be possible to stay below 10MB.

Adding CMake options to the build script isn't really that much and it's a one time thing.

Andrey1994:
It's just 20MB for all platforms and also it can be separated by OS if at a certain point these files will be too big. 20MB is literally nothing and it's way easier to download it than wait for compilation. 

Numpy whl is 14MB, TensorFlow whl for CPU is almost 500MB for example and everybody is using it as a precompiled binary, even in Tensorflow for Rust they download binaries by default(https://github.com/tensorflow/rust#usage), 10MB limitation in a package registry is very weird as for me

PR #405, opened 2022-01-03
Add muse2016 (native and BLED) support

JohnGriffiths:
Very keen to get this addition into brainflow master, and switch eeg-notebooks over to brainflow as default backend for all muse devices asap. 

I am very naive as to brainflow's inner workings however, so apologies for the limits in my current understanding of what's needed and how to implement. 

Following @Andrey1994 's recommendations I've followed all the items on the [dev instructions](https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#instructions-to-add-new-boards-to-brainflow) for adding two new boards: 'MUSE_2016' and 'MUSE_2016_BLED'. 

This was fairly straightforward as most of it was simply copying from the corresponding Muse 2 code lines. 
Need to double check some things like channel numbers for gyro etc., but I believe the numbers in this PR are correct. 

So the easy part of adding these two new boards is I think done or nearly done. The hard part I think will involve some additions / modifications to the [muse bglib](https://github.com/brainflow-dev/brainflow/tree/master/src/board_controller/muse) code, but I am unsure how to determine what is needed and generally how to proceed there. 

Please advise!

Thanks!


Andrey1994:
@JohnGriffiths I've made some minor changes, could you test them? I have no muse2016

PR #415, opened 2022-01-27
add link to compilation in docs

evaesteban:


Andrey1994 (changes requested):
use https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html and also there are more places that refer to a compilation of the source code(python binding, java binding, etc)

Issue #245, opened 2021-02-25
add unit tests for c++ code


Andrey1994:
We have plenty of integration tests, but it would be good to add unit tests to test small functions as well

PR #429, opened 2022-04-04
Add bluetooth unit tests

philippitts:
Relates to: https://github.com/brainflow-dev/brainflow/issues/245

Introduces [Google Test](https://github.com/google/googletest) framework and provides unit tests for the Bluetooth utility functions. Tests may be optionally built with the `BUILD_TESTS` CMake flag (off by default). Enabling this flag will also enable flags to download and build Google Test. Disabling the flag will remove the Google Test CMake options. Also introduces "Run unit tests" job to Windows and Unix action workflows.

Notable changes to source code (beyond tests, build scripts, and CI) include:
- Moved C++ integration tests to `test/cpp/demos/` folder. This was done so that the source file structure can be replicated in the tests folder. CI scripts have been updated accordingly. If these changes are not wanted, the corresponding commit can be dropped.
- `bluetooth_get_data` and `bluetooth_write_data` do not return error codes, and instead return 0 on failure. This was done because these functions return the number of processed bytes on success so the return was ambiguous.

### GitHub Actions Job
image

### CMake GUI - BUILD_TESTS Not Selected
image

### CMake GUI - BUILD_TESTS Selected
image


Andrey1994:
Hi!
I like everything except this change 
```bluetooth_get_data and bluetooth_write_data do not return error codes, and instead return 0 on failure. This was done because these functions return the number of processed bytes on success so the return was ambiguous.```

Its common to return negative values from methods like `recv` or `send` in case of errors and currently it does exactly this. This enum values which we return right now are negative and I think its better than returning 0 in case of error

Andrey1994 (changes requested):
everything is fine, I wrote my only concern in previos comment

Andrey1994:
and one more note: could you also add a parameter to tools\build.py? Its a small python wrapper to dont type cmake commands and arguments manually

philippitts:
> Hi! I like everything except this change `bluetooth_get_data and bluetooth_write_data do not return error codes, and instead return 0 on failure. This was done because these functions return the number of processed bytes on success so the return was ambiguous.`
> 
> Its common to return negative values from methods like `recv` or `send` in case of errors and currently it does exactly this. These enum values which we return right now are negative and I think its better than returning 0.
> 
> Example: https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-sendto#return-value (socket_error is -1 in fact)

I think the OS functions will return negative error codes as you described, but if the device is not already created, the functions in `bluetooth_functions` will return a `DEVICE_IS_NOT_CREATED_ERROR` which converts to a positive int (7). This is the part I think is ambiguous because if the device is not created, client code may think it has received 7 bytes.

Would it make sense make the assignments in the `SocketBluetoothReturnCodes` enum negative? Along those lines, would it make sense to generalize the error codes in `SocketBluetoothReturnCodes` to general errors that are not platform specific? For example, `WSA_STARTUP_ERROR` and other `WSA_*` codes will only be returned on Windows., and the purpose of `bluetooth_functions` seems to be to create a platform independent layer. There's a tradeoff here between error detail and platform independence.

image

Andrey1994:
ah, yes, I thought they are already negative. lets fix it

<<For example, WSA_STARTUP_ERROR and other WSA_* codes will only be returned on Windows, and the purpose of bluetooth_functions seems to be to create a platform independent layer. There's a tradeoff here between error detail and platform independence.

yes, its fine

Andrey1994:
merge master branch and it will fix failed pipelines

philippitts:
> merge master branch and it will fix failed pipelines

Thanks - pipelines are passing now and the changes you requested are done. I also added a commit that moves all the demos to example folders in their relative hierarchies. I thought this would help clarify what can be used as a usage example and what is used exclusively for tests (e.g. with assertions). If you think that makes sense, I can squash the previous commit into it. If you don't like it, I'll just drop the last commit and we should good to go.

Andrey1994:
thanks, looks good.

I dont like the part about examples vs tests. There are still some examples\tests in binding folders(like this https://github.com/philippitts/brainflow/tree/add-bluetooth-unit-tests/python-package/examples or https://github.com/philippitts/brainflow/tree/add-bluetooth-unit-tests/java-package/brainflow/src/main/java/brainflow/examples)

Its fine to move them all to a single top level folder, but I think it should be done for all of them at once and in its own PR.

philippitts:
Makes sense to me. I'll drop that commit and resubmit.

Andrey1994 (approved):

PR #438, opened 2022-05-19
Fixing Matlab binding for get_csp method

dvaposto:


matlab-package/brainflow/DataFilter.m
```diff
@@ -193,11 +193,11 @@ function disable_data_logger()
         function [filters, eigenvalues] = get_csp(data, labels)
             % get common spatial patterns
             task_name = 'get_csp';
-            n_epochs = size(data, 1);
-            n_channels = size(data, 2);
-            n_times = size(data, 3);
+            n_epochs = size(data, 3);
```
Andrey1994, line 196: as far as I understand before the order was epochs, channels, times and it matched other languages(at least c# https://github.com/brainflow-dev/brainflow/blob/master/csharp-package/brainflow/brainflow/data_filter.cs#L338 ). Here its changed and I think it should be consistent. The same about julia I guess. 

also, lets add matlab test to https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/run_matlab.yml

Andrey1994 (changes requested):

Andrey1994 (approved):

Andrey1994:
looks good, thanks

Issue #517, opened 2022-07-04
[Java] Can there be a solution that is better than `BoardShim.unpack_from_jar`?


mgroth0:
It is not ideal that .dylib files are appearing in the project folder on `gradle run`. Can there be an option to use another location?

[This doc](https://java-native-access.github.io/jna/4.2.1/com/sun/jna/NativeLibrary.html#library_search_paths) gives some useful information for different places that these files can be. "Context class loader classpath" seems like a good option.

Andrey1994:
Hi, idk, I didnt find a better option when I was writting it.
If context loader classpath works for you feel free to send a PR, it will be great. Another approach to try can be to change env variables

Andrey1994:
you need to compile it first using build.py https://brainflow.readthedocs.io/en/stable/BuildBrainFlow.html#compilation-of-core-module-and-c-binding 

And as far as I understand you need to copy files to subfolder like linux-x64, etc https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow/pom.xml#L50. Maybe you will need to change files like build.cmake(there are a lot of them, you probably need to run smth like `find . -name build.cmake` https://github.com/brainflow-dev/brainflow/blob/master/src/board_controller/build.cmake#L172

Andrey1994:
Also, when you compile it locally you do it only for one OS, while in jar file for releases there are libs for all OSes so this jar is x-platform. Its done by this CI job https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/deploy.yml, it copies files for all OSes from AWS

Andrey1994:
completed

PR #519, opened 2022-07-06
make dylibs go to temp folder / use JNI context classloader on mac

mgroth0:
Resolves https://github.com/brainflow-dev/brainflow/issues/517 for mac. Should be generalized to all platforms.

This PR changes the way that the native libraries are built into the java package on mac. It also contains a commit adding a few lines to the .gitignore because these files were generated in my project while I was developing the PR and I figured they belonged in the .gitignore.

I have tested the jar in a simple program that connects to an OpenBCI Ganglion, and it works. No more dylibs are copied into the project folder.

I think that generalizing this to other platforms will be straightforward:

1. Create more `resource` nodes in `pom.xml` for each platform. The correct platform name needs to be used. replace `darwin-aarch64` with the result of [this function](https://sourcegraph.com/github.com/java-native-access/jna@999c60a1fa0ab1154b07edffa3748ce46d8d3b89/-/blob/src/com/sun/jna/Platform.java?L307:19&subtree=true) for each platform.
2. In `Boardshim.java`, `copy_to_temp_dir` should be generalized to other platforms by getting the correct temp dir from [this logic](https://github.com/java-native-access/jna/blob/master/src/com/sun/jna/Native.java#L1321). Currently it only has the correct temp dir for Mac.
3. In the `static` block, `unpack_from_jar` should eventually be replaced by `copy_to_temp_dir` for all platforms. 
4. Each platform should be tested to make sure all native libraries are correctly loaded and no native libraries are copied into the project folder any more.


Andrey1994 (changes requested):
Hi and thanks for PR.

First of all it should be done for all 3 main classes, not only for BoardShim. There are also DataFilter and MLModule.

Also, I think it should be done for all OSes at the same time and it should be consistent.

I am also not sure in folder name "darwin-aarch64", maybe it doesnt work for intel based macs. Libraries for brainflow are universal and support both architectures.

According to this doc https://java-native-access.github.io/jna/4.2.1/com/sun/jna/NativeLibrary.html#library_search_paths "Context class loader classpath. Deployed native libraries may be installed on the classpath under ${os-prefix}/LIBRARY_FILENAME, where ${os-prefix} is the OS/Arch prefix returned by [Platform.getNativeLibraryResourcePrefix()](https://java-native-access.github.io/jna/4.2.1/com/sun/jna/Platform.html#getNativeLibraryResourcePrefix--). If bundled in a jar file, the resource will be extracted to jna.tmpdir for loading, and later removed (but only if jna.nounpack is false or not set)." if I understand it correctly you should not extract it by yourself, jna should unpack it automatically, so more likely we dont even need copy_to_temp_dir method and JNA can(should) do it for all OSes

Finally you need to use this formatter https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow_formatter.xml to keep code style consistent

Andrey1994:
> > Also, I think it should be done for all OSes at the same time and it should be consistent.
> > I am also not sure in folder name "darwin-aarch64", maybe it doesnt work for intel based macs. Libraries for brainflow are universal and support both architectures.
> 
> To solve the intel mac problem, we can just copy the native libraries to both architecture folders in the `pom.xml`. I don't know what the other architecture names are. I think the best way to find out would be to iterate through all platforms somehow and run `com.sun.jna.Platform.getNativeLibraryResourcePrefix()` and record the return value for each system. That function is also package private so [reflection will be needed to access it](https://www.geeksforgeeks.org/how-to-call-private-method-from-another-class-in-java-with-help-of-reflection-api/).

yes, I was also thinking about copying to a folder with different name. There is a project that does smth similar. You can download this jar and open it with 7zip https://mvnrepository.com/artifact/com.microsoft.onnxruntime/onnxruntime/1.11.0 it will tell you at least some of the folder names used by jna

Another options which is easier than reflection is to take a look at the source code https://github.com/java-native-access/jna/search?q=getNativeLibraryResourcePrefix and https://github.com/java-native-access/jna/blob/bf60e51eace6dffa18548019e2ba398ff84904ef/src/com/sun/jna/Platform.java#L307

java_package/brainflow/src/main/java/brainflow/BoardShim.java
```diff
@@ -152,31 +152,14 @@ int get_current_board_data (int num_samples, double[] data_buf, int[] returned_s
         } else
         {
             // need to extract libraries from jar
-            unpack_from_jar (lib_name);
+            /* unpack_from_jar (lib_name); */
```
Andrey1994, line 155: why is it commented out?

java_package/brainflow/src/main/java/brainflow/NativeLoader.java
```diff
@@ -0,0 +1,44 @@
...
+        {
+            System.err.println ("file: " + lib_name + " is not found in jar file");
+            return null;
+        }
+    }
+
+    static void copy_to_temp_dir (String lib_name)
+    {
+        // https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html
+        File jnatmp = new File (System.getProperty ("user.home"), "Library/Caches/JNA/temp/" + lib_name);
```
Andrey1994, line 32: I dont like hardcode for jna temp dir, there is a method for it https://github.com/java-native-access/jna/blob/bf60e51eace6dffa18548019e2ba398ff84904ef/src/com/sun/jna/Native.java#L1321 and its OS agnostic

Andrey1994:
fixed in another PR for all OSes

PR #528, opened 2022-07-28
fixing read_file issue with invalid input

Andrey1994:


Andrey1994:
fixed not a number in the file and invalid size for rows

PR #542, opened 2022-09-05
add support for mentalab devices

Andrey1994:

Tested on windows, can connect to device and get data.

Todo and issues:
- [x] can connect to device only when LED is blue and blinking, after some time it gets green and you have to restart the device
- [x] temperature data looks completely wrong, its a random value between 20 and 120, need to check parsing
- [ ] accel and gyro data, instead expected values like -1..0..1(these units used in brainflow) I get something I dont understand, the same in explorepy, so its not an issue with parsing
- [x] add docs about explore 4 and 8 channels boards
- [x] add release notes to website
- [ ] need to test on linux and macos, have no hardware
- [ ] cannot test correctness of eeg parsing(I dont get how to connect these electrodes), dont understand the idea of 5th channel... is it a reference which should be subtracted from other channels? these value does not change for me 
- [x] it seems like fw supports timestamp syncronization, need to implement it or at least use sampling rate to correct timestamps which come from batches of datapoints
- [ ] need to test 8 chan board, have no device 

To do later in case of requests:

- add support for sending commands(at least for sampling rate and maybe on\off)
- add support for impedance measurement


Andrey1994:
[data_env.csv](https://github.com/brainflow-dev/brainflow/files/9485933/data_env.csv)
[data_orn.csv](https://github.com/brainflow-dev/brainflow/files/9485934/data_orn.csv)
[data_eeg.csv](https://github.com/brainflow-dev/brainflow/files/9485935/data_eeg.csv)

data files from explore.py example attached, still dont understand values there

Andrey1994:
Update:

its not a parsing issue in brainflow for accelerometer, using explorepy I get the same values:

```
accelerometer
[ -19.093 -134.627 1011.746]
```

Andrey1994:
temperature is also correct, I looked at the wrong column before

Andrey1994:
fixed linux issue, fixed parsing, added commands for sampling rate and test signal. validated against explorepy using test signal

Issue #555, opened 2022-10-01
Move synthetic data streams that would have lower sampling rates to SyntheticAux


philippitts:
**Use case:**
Emulating devices with synthetic or algorithmic data is useful in the development of code that consumes Brainflow. Many boards group data streams with slower sampling rates into the auxiliary and/or ancillary board variations. Currently data streams that often have slow sampling rates are in the default Synthetic Brainflow board definition, but it would make more sense for them to be in the auxiliary variation. Doing this would allow synthetic devices to better mirror real ones, and subsequently would allow client code to better abstract synthetic and real devices.

**Describe the solution you'd like:**
Move the battery, eda, ppg, accel, gyro, resistance and temperature channels to the SyntheticAux board definition.

Andrey1994:
makes sense but moving may break playback of synthetic board for already recorded files and some existing user code which currently gets it from default preset.
probably instead moving it can be copied to auxiliary preset with smaller sampling rate

Andrey1994:
and its ok if some data types are duplicated between presets, especially for synthetic board

PR #556, opened 2022-10-01
Copy slow synthetic data streams to aux board

philippitts:
Closes https://github.com/brainflow-dev/brainflow/issues/555

### DIFF
--- a/src/board_controller/brainflow_boards.cpp
+++ b/src/board_controller/brainflow_boards.cpp
@@ -77,17 +77,17 @@ BrainFlowBoards::BrainFlowBoards()
         {"name", "SyntheticAux"},
         {"sampling_rate", 250},
         {"package_num_channel", 0},
-        {"timestamp_channel", 5},
-        {"marker_channel", 6},
+        {"timestamp_channel", 18},
+        {"marker_channel", 19},
         {"num_rows", 20},
-        {"other_channels", {1, 2, 3, 4}},
-        {"battery_channel", 7},
-        {"accel_channels", {8, 9, 10}},
-        {"gyro_channels", {11, 12, 13}},
-        {"eda_channels", {14}},
-        {"ppg_channels", {15, 16}},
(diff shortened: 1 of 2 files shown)


src/board_controller/brainflow_boards.cpp
```diff
@@ -79,8 +79,15 @@ BrainFlowBoards::BrainFlowBoards()
         {"package_num_channel", 0},
         {"timestamp_channel", 5},
         {"marker_channel", 6},
-        {"num_rows", 7},
-        {"other_channels", {1, 2, 3, 4}}
+        {"num_rows", 20},
+        {"other_channels", {1, 2, 3, 4}},
```
Andrey1994, line 83: its better to add other channels to the end, they are meaningless most of the time and when you store data in a file they will be in the beginning but will not be interesting for the users.

Also, for all boards timestamp and marker channels are in the end, lets keep it consistent

Andrey1994 (changes requested):
I think it will not work, in synthetic_board.cpp you need to fill it properly

philippitts:
> I think it will not work, in synthetic_board.cpp you need to fill it properly

Resolved in https://github.com/brainflow-dev/brainflow/commit/0ab043c898871850060512a4b349aca82a9d8c4e

src/board_controller/brainflow_boards.cpp
philippitts: Resolved in https://github.com/brainflow-dev/brainflow/pull/556/commits/81ac4568e73e8a23eba69f37067a4fa5ee6f4649

philippitts:
Assuming these fixes look okay, let me know if you'd like me to rebase and combine these commits into one before submitting for a merge.

Andrey1994:
there is squash and merge option in github to create single commit, rebasing is not needed

Andrey1994 (approved):

PR #562, opened 2022-10-14
Add cyton gain tracking test

retiutut:
- [x] set gain to 1 for channel 1 and confirm that data is not 24 times bigger\smaller

- [x] try to set it for channel 9 for cyton wo daisy and confirm error

- [x] try to set it for channels 1 and 9 for cyton using single call of config_board method and see what will happen

- [x] try to send smth like config_board('x3020 000Xgarbagehere')

Try more test cases if needed.


Andrey1994:
I am not sure that we need an integration test for that in the repo, probably will be enough to run once and forget about it. For some auto checks unit tests suits better here but its optional

Andrey1994:
also I think you should do it once at a time, otherwise it will be impossible to validate data

Andrey1994:
I think the easiest check will be to calc mean of the data in 5 seconds before config board call and in 5 seconds after, no need psd_welsch method

Andrey1994:
and need to compare behavior against upstream version during these tests, ideally to have smth like a table with mean values after different config commands for old version and in this pr

retiutut:
@Andrey1994 I've written the requested tests and they can be reviewed. I'm seeing some gaps in the code as a result. I've noticed that setting channel 9 when Cyton board is being used without Daisy attached does not return an error when it should.

Andrey1994:
looks ok at the first sight. what are the outputs for each of them for current version and using this PR?

Andrey1994:
about channel 9 for cyton, its only for channel number 9 right? for 10 should be ok. Could you change this line https://github.com/brainflow-dev/brainflow/pull/546/files#diff-e6ed7b7c41e69e9675da38702d00d7fbdbea0406ea1a0c1bb80466fb2d353727R63 to `>=` and try again?

PR #565, opened 2022-10-18
Fixed a an issue with memory being written to after it was freed, leading to abort signals from malloc

bokbok:
There was an issue while calling the config_board method from the rust library, leading to SIGABRT killing the process. This was a bit hard to track down as exactly when the abort happened wasnt deterministic. It could be at any time a memory allocation via malloc took place after config_board was called. The abort was due to the allocator detecting that memory had been written to after it was freed  Ultimately it was determined that the c(++) code that was called was performing a strcpy in to response CString that was passed to this code. This copy was resulting in the memory being written to after it was freed.

If you refer to the doco for CString.into_raw() you will see "The C side must not modify the length of the string (by writing a null somewhere inside the string or removing the final one) before it makes it back into Rust using [CString::from_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw). See the safety section in [CString::from_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw).". The code here as it was was breaking this constraint. I am not 100% sure if this was causing the memory to be written to after it was freed or not, but using a preallocated character buffer whose lifecycle was managed on the stack seems to have fixed the issue.

### DIFF
--- a/rust_package/brainflow/src/board_shim.rs
+++ b/rust_package/brainflow/src/board_shim.rs
@@ -218,27 +218,25 @@ impl BoardShim {
     pub fn config_board<S: AsRef<str>>(&self, config: S) -> Result<String> {
         let config = CString::new(config.as_ref())?;
         let mut response_len = 0;
-        let mut c_string: [c_char; 8192] = [0; 8192];
+        let mut result_char_buffer: [c_char; 8192] = [0; 8192];
 
         let config = config.into_raw();
         let (res, response) = unsafe {
             let res = board_controller::config_board(
                 config,
-                c_string.as_mut_ptr(),
+                result_char_buffer.as_mut_ptr(),
                 &mut response_len,
                 self.board_id as c_int,
                 self.json_brainflow_input_params.as_ptr(),
             );
             let _ = CString::from_raw(config);
-            let resp = CStr::from_ptr(c_string.as_ptr());
+            let resp = CStr::from_ptr(result_char_buffer.as_ptr());
 
             (res, resp)
         };
         check_brainflow_exit_code(res)?;
         Ok(response
             .to_str()?
-            .split_at(response_len as usize)
-            .0
             .to_string())
     }
 
@@ -359,34 +357,30 @@ pub fn log_message<S: AsRef<str>>(log_level: LogLevels, message: S) -> Result<()
 
 /// Get board description as json.
 pub fn get_board_descr(board_id: BoardIds, preset: BrainFlowPresets) -> Result<String> {
+    const MAX_CHARS: usize = 16000;
     let mut response_len = 0;
-    let response = CString::new(Vec::with_capacity(16000))?;
-    let response = response.into_raw();
+    let mut result_char_buffer: [c_char; MAX_CHARS] = [0; MAX_CHARS];
     let (res, response) = unsafe {
-        let res = board_controller::get_board_descr(board_id as c_int, preset as c_int, response, &mut response_len);
-        let response = CString::from_raw(response);
+        let res = board_controller::get_board_descr(board_id as c_int, preset as c_int, result_char_buffer.as_mut_ptr(), &mut response_len);
+        let response = CStr::from_ptr(result_char_buffer.as_ptr());
         (res, response)
     };
     check_brainflow_exit_code(res)?;
-    Ok(response
-        .to_str()?
-        .split_at(response_len as usize)
-        .0
-        .to_string())
+    Ok(response.to_str()?.to_string())
 }
 
 /// Get names of EEG channels in 10-20 system if their location is fixed.
 pub fn get_eeg_names(board_id: BoardIds, preset: BrainFlowPresets) -> Result<Vec<String>> {
+    const MAX_CHARS: usize = 16000;
     let mut response_len = 0;
-    let response = CString::new(Vec::with_capacity(16000))?;
-    let response = response.into_raw();
+    let mut result_char_buffer: [c_char; MAX_CHARS] = [0; MAX_CHARS];
     let (res, response) = unsafe {
-        let res = board_controller::get_eeg_names(board_id as c_int, preset as c_int, response, &mut response_len);
-        let response = CString::from_raw(response);
+        let res = board_controller::get_eeg_names(board_id as c_int, preset as c_int, result_char_buffer.as_mut_ptr(), &mut response_len);
+        let response = CStr::from_ptr(result_char_buffer.as_ptr());
         (res, response)
     };
     check_brainflow_exit_code(res)?;
-    let names = response.to_str()?.split_at(response_len as usize).0;
+    let names = response.to_str()?;
 
     Ok(names
         .split(',')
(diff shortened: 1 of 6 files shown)


bokbok:
Oh and FYI this was happening on an intel mac, macos 12.1

Andrey1994 (commented):
Hi and thanks!

I have a couple of questions(not an expert in rust at all), is it ok to return a pointer to the object allocated on stack in rust? Or cstr::from_char_ptr handles it fine and does smth like memcpy? In C it will lead to segfault. 

Did you consider the alternative to declare it as smth like array of unsigned chars? If its an array I guess its ok to write null term char anywhere and convert it to string later.

bokbok:
> Hi and thanks!
> 
> I have a couple of questions(not an expert in rust at all), is it ok to return a pointer to the object allocated on stack in rust? Or cstr::from_char_ptr handles it fine and does smth like memcpy? In C it will lead to segfault.
> 
> Did you consider the alternative to declare it as smth like array of unsigned chars? If its an array I guess its ok to write null term char anywhere and convert it to string later.

In an unsafe block yes its ok as it isnt transferring ownership to the called method and the array lives as long as the calling method does.

I am not sure if cstr::from_char_pointer performs a copy or not, but this object only lives as long as the method that creates it. The conversion to a rust string definitely copies the character array as rust represents its strings differently.

Andrey1994:
Got that, thanks!

There are a few more methods which do it the same way:

* get_device_name
* get__eeg_channles_names
* get_board_descr
* get_version(in boardshim, datafilter and mlmodel)

Could you update it there also?

bokbok:
> 

yeah, sure

bokbok:
@Andrey1994 I have pushed changes with fixes to the other identified functions. These definitely caused SIGABORTs on my intel mac also. I have supplied tests for all the modified methods. It would be good to get these tests running in CI/CD. It would be good to get a proper rust crate published to the package registry. I'm happy to help out to get this done.

Andrey1994:
Thanks! I will take a look later today.

Regarding rust package, I was going to implement it but faced the issue that it doesnt allow to publish packages bigger than 20mb and all compiled Cpp libs are bigger than that.

Regarding CI/CD that would be good, we have some integration tests for rust in github actions. I believe its splitted between run_windows and run_unix workflows. You can add more tests there, and currently we do not have unit tests. Integration tests are also used as code samples at brainflow docs, so its why we prioritized them first

rust_package/brainflow/src/board_shim.rs
```diff
@@ -518,3 +513,70 @@ gen_vec_fn!(
...
+        #[test]
+        fn test_get_eeg_names() {
+            assert_eq!(vec!["Fz", "C3", "Cz", "C4", "Pz", "PO7", "Oz", "PO8", "F5", "F7", "F3", "F1", "F2", "F4", "F6", "F8"],
+                       get_eeg_names(BoardIds::SyntheticBoard, BrainFlowPresets::DefaultPreset).unwrap());
+
+        }
+
+        #[test]
+        fn test_get_board_descr() {
+            let expected = "{\"accel_channels\":[17,18,19],\"battery_channel\":29,\
```
Andrey1994, line 543: dont like this check. it probably should look like comparison with json object by keys and values not as string comparison. And for other languages `get_board_descr` returns dicts\jsons\class instances instead string, idk how to do the same in rust but should be possible

Andrey1994 (commented):
code looks fine, need to do some minor fixes with tests and will be good to go

rust_package/brainflow/src/board_shim.rs
bokbok: I wrote the test reverse engineering existing functionality, using the synthetic board. It was really just enough to see that I hadnt broken existing functionality in that context (and also allowed my to prove that multiple calls caused SIGABRT). I dont really like that its bound up with a config that is coming from the C++ layers and is way out of this test. I'd probably rather test just a simple field or two here which are unlikely to change. 

Does your comment mean that you would rather we should change the contract of the function to return a json object like structure? I think its probably a good idea to do something like that, but as rust is statically typed its probably best to map it on to some types representing the the config instead of returning some sort of generic json object which is difficult to navigate. (Python etc are a lot easier/simpler for json conversions)

rust_package/brainflow/src/board_shim.rs
```diff
@@ -518,3 +513,70 @@ gen_vec_fn!(
...
+#[cfg(test)]
+mod tests {
+    #[cfg(test)]
+    mod functions {
+        use crate::board_shim::{get_version, get_device_name, get_eeg_names, get_board_descr};
+        use crate::{BoardIds, BrainFlowPresets};
+
+        #[test]
+        fn test_it_gets_the_version() {
+            assert_eq!("0.0.1", get_version().unwrap());
```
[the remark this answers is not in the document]
bokbok: I think a regex is fine if we cant really predict the actual value

bokbok:
> 

How about bundling the c++ source with the rust code and compiling it as part of the crate installation? This may be problematic with windows though...  I assume the c++ code isnt currently distributed via package managers as a sharedlib?

Andrey1994:
Its distributed as shared libs as a part of all bindings, I wanted to do the same with rust but didnt find a way.. for python they will be directly in whl file which you install, the same for c# and java 

There is also a conanfile but its not pushed to conan registry

If its still impossible then probably the only option is to build cpp from rust, the main requirement for it - dont add its own build system and scripts, it should use existing cmakefile and we should not update anything in rust when we make changes on cpp side

bokbok:
> 

How many architectures and OSes are you providing shared libs for? Are things like libc statically linked? It feels like providing shared libs as part of the bindings might be less sustainable than providing source to build the shared libs.

bokbok:
Those tests have been changed to do some simple regex matching making them more brittle to change. I left any refactoring of the result for get_board_desc for another exercise. I'm happy to make further contributions to the rust code when I have time..

bokbok:
> 

Scrub that I missed the other two get_version tests

Andrey1994:
> > 
> 
> How many architectures and OSes are you providing shared libs for? Are things like libc statically linked? It feels like providing shared libs as part of the bindings might be less sustainable than providing source to build the shared libs.


https://brainflow.readthedocs.io/en/stable/BuildBrainFlow.html#precompiled-libraries-in-package-managers-nuget-pypi-etc

it covers almost all possible platforms, for macos its a universal binaries for both x86 and arm, with min supported version 10.15, for linux it should work everywhere with x86 because of manylinux containers, for windows all x86 starting from 8.1

there is a need to build only for linux\windows + arm and its not popular use case, also I am going to add it later

Building from source is challenging for most of the users, especially for python\matlab. They are not familiar with cmake quite often and there are plenty of compilation options and flags(check arguments in tools/build.py)

bokbok:
> > > 
> > 
> > 
> > How many architectures and OSes are you providing shared libs for? Are things like libc statically linked? It feels like providing shared libs as part of the bindings might be less sustainable than providing source to build the shared libs.
> 
> https://brainflow.readthedocs.io/en/stable/BuildBrainFlow.html#precompiled-libraries-in-package-managers-nuget-pypi-etc
> 
> it covers almost all possible platforms, for macos its a universal binaries for both x86 and arm, with min supported version 10.15, for linux it should work everywhere with x86 because of manylinux containers, for windows all x86 starting from 8.1
> 
> there is a need to build only for linux\windows + arm and its not popular use case, also I am going to add it later
> 
> Building from source is challenging for most of the users, especially for python\matlab. They are not familiar with cmake quite often and there are plenty of compilation options and flags(check arguments in tools/build.py)

Right I can see that being difficult, although a lot of python libs do come with native deps and seamlessly work. I dont know anything about matlab but I get that the userbase are often not from a background where setting up toolchains are easy. 

The stuff I am building I am intending to move to raspberry pi or nvidia jetson, so I would argue ARM is important for linux ;)

bokbok:
I have performed all changes now, please review and merge if happy

Andrey1994 (approved):

rust_package/brainflow/src/board_shim.rs
Andrey1994: >Does your comment mean that you would rather we should change the contract of the function to return a json object like structure? I think its probably a good idea to do something like that, but as rust is statically typed its probably best to map it on to some types representing the the config instead of returning some sort of generic json object which is difficult to navigate. (Python etc are a lot easier/simpler for json conversions)

yes, smth like this. In java\c# I did it using templates instead json string. https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow/src/main/java/brainflow/examples/BandPower.java#L26 https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow/src/main/java/brainflow/BoardDescr.java  https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow/src/main/java/brainflow/BoardShim.java#L583

Andrey1994: its ok to do it in another pr, no need to change now

PR #566, opened 2022-10-19
Add fixes to gain tracker

retiutut:
Tested using scripts found here: https://github.com/retiutut/brainflow/tree/gain_tracker-rw/python_package/examples/tests


Andrey1994:
need to remove iostream and all debug prints. In release builds for production there should not be debug dev messages directly in stdout

Andrey1994 (approved):

PR #570, opened 2022-10-22
add set_index_percentage for playback board and fix synthetic board

### LINKED ISSUE #567: Add configuration option for start index in playback boards
**Use case:**
There are several configuration options for the Playback File Board that can be set via the `config_board` method. It would be useful to have another option that sets the index in the playback file to start playback from. This would enable scrubbing through the file and starting the file a specific locations (e.g. an start index found using a marker for alignment with other modalities).

**Describe the solution you'd like:**
Add an option to the `config_board` method that supports starting the playback board at a specified sample index in the file.

**Describe alternatives you've considered:**
This can be done outside of Brainflow by reading the file and sending data to the DataFilter interfaces, but doing so replicates much of the functionality already implemented in the Playback File Board. Alternatively, a version of the file could be edited and saved so that it only contains data starting from the desired index, but this is inefficient.

Andrey1994:

closes #567 

it allows to set a percentage instead absolute value. Percentage is better because for absolute values also need to design the api to get max value for each preset and api to set index should be per preset(playbackfile) while using percentage we can set index for all presets at once and do not need to add api to get size of playback file.


Andrey1994:
it also doesnt preload entire file into memory which is pretty cool. But it stores offsets for each line and its limited by the size of `long int` and consumes some memory

Andrey1994:
when you use this method keep in mind that it's up to you to clean old data using `get_board_data` otherwise you will see gaps in the data stream. Also when you set a position it will look like a new stream starting from the specified offset, it means if you plot data in a window it may take a few seconds

PR #572, opened 2022-10-25
Update Java jna dependency to 5.12.1

retiutut:


Andrey1994:
Does it work on latest macos? And did you confirm that old version doesnt? 

Also need to update *very carefully* jnilibs for android to match the version and abi it should be placed in the correct abi folder https://github.com/brainflow-dev/brainflow/blob/master/tools/jniLibs/README.md

PR #574, opened 2022-10-30
add support for ganglion via native ble

### LINKED ISSUE #554: update simpleble

### LINKED ISSUE #568: Create BLELibBoard for Ganglion devices on MacOS 10.15+
**Use case:**
There is a problem with newer MacOS devices using Apple silicon processors that causes data to stream slowly into BrainFlow and client software. This problem appears to be related to changes in the USB interfaces for devices running on Apple silicon. Creating a BLELibBoard will circumvent the issue and allow data to stream at normal sampling rates.

**Describe the solution you'd like:**
Implement a BrainFlowBoard that derives from BLELibBoard and supports streaming Ganglion data across BLE.

**Describe alternatives you've considered:**
The problem seems to be related to USB interfaces. Several variations of USB adapters ranging from USB 2.0, USB 3.0, and USB 3.1 have been tried. Powered and unpowered adapters have also been tried. Different adapters will stream at slightly different speeds, but no adapters will stream at the full 200Hz sampling rate. The problem appears on both old and new combinations of Ganglion devices and dongles. The issue seems to be related to either a MacOS update or possibly a firmware update for USB devices on MacOS. The issue does not appear on MacOS devices not running Apple silicon or on other operating systems. It also does not appear on MacOS devices with USB A ports (only USB C ports that require an adapter).

Andrey1994:
closes #554 
closes #568


Andrey1994:
image

PR #581, opened 2022-11-22
Fix ganglion native ble

retiutut:


Andrey1994:
need to confirm that it still works fine on windows and check on macos, if there are still issues need to try to update simpleble one more time(there is ongoing work for v 0.6 which is not released yet but merged to main branch)

Andrey1994:
this one looks important https://github.com/OpenBluetoothToolbox/SimpleBLE/commit/50d2e39353ab6cacb09c7dc24834e1c1c2070439

Andrey1994:
I've updated simpleble to unreleased version, it may be dangerous a little since I have no idea how stable it is. Need to confirm that it works now on all OSes

PR #596, opened 2023-01-17
Implement ICA

Andrey1994:


Andrey1994:
todo list:

* [ ] test that it makes sense and calculates smth useful correctly(need help), also need to validate against R fastICA package
* [ ] understand how to remove some components and add required methods for it
* [ ] update all bindings, for now its only for python

PR #600, opened 2023-01-25
Use master_board parameter for csharp streaming and playback

philippitts:
The C# bindings currently use the `BrainFlowInputParams::other_info` member to set master board ID for streaming and playback boards. In other bindings (e.g. python) this has been updated to use the the `BrainFlowInputParams::master_board` member. This PR updates the C# binding to match that pattern.


Andrey1994:
seems like I forgot to update it for c#

PR #624, opened 2023-05-07
add reference_range param

furmanlukasz:
add control to set amplifier reference range. Accessible values are 1.0, 0.75, 0.15 Volt. I also change the default sampling rate from 2000Hz to 512Hz.

### DIFF
--- a/src/board_controller/ant_neuro/ant_neuro.cpp
+++ b/src/board_controller/ant_neuro/ant_neuro.cpp
@@ -62,11 +62,21 @@ AntNeuroBoard::AntNeuroBoard (int board_id, struct BrainFlowInputParams params)
     stream = NULL;
     try
     {
-        sampling_rate = board_descr["sampling_rate"];
+        sampling_rate = board_descr["default"]["sampling_rate"];
     }
     catch (...)
     {
-        sampling_rate = 2000;
+        sampling_rate = 512;
+    }
+
+     // Initialize "reference_voltage" with a default value or from the board descriptor if available
+    try
+    {
+        reference_range = board_descr["default"]["reference_range"];
+    }
+    catch (...)
+    {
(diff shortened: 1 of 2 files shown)


Andrey1994 (changes requested):
hi and thanks for this PR!

Overall, it looks fine. But some changes are needed: setup clang format formatter as described here https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style and reformat your code, it will fix the failed CI job.

Also, I didnt like this line `reference_range = board_descr["default"]["reference_range"];` 
it will also fail with exception, since its not defined in json. And it should not be defined in json in fact since this json file is just an internal helper for high order methods like `BoardShim.get_sampling_rate`, etc.

I think you can just set the initial reference_range to 1.0 in constructor, or even better set it to the one of the supported values and get the list of supported values using ant neuro api.

Also, thanks for adding missed `["default"]` seems like I forgot to add it for ant neuro boards and I even do not fully understand how it was working before this change

src/board_controller/ant_neuro/ant_neuro.cpp
```diff
@@ -297,7 +309,26 @@ int AntNeuroBoard::config_board (std::string config, std::string &response)
...
+            new_reference_range = std::stod (value);
+        }
+        catch (...)
+        {
+            safe_logger (spdlog::level::err, "format is '{}value'", rv_prefix.c_str ());
+            return (int)BrainFlowExitCodes::INVALID_ARGUMENTS_ERROR;
+        }
+
+        // Set the reference_voltage value, assuming you have a class member called reference_voltage
+        reference_range = new_reference_range;
```
Andrey1994, line 327: For sampling rates we check that provided value is correct using GetSamplingRatesAvailable method(check above), is there smth like that for reference voltage? If yes, lets add it here too

Issue #619, opened 2023-03-26
add method to rereference channels


Andrey1994:
https://pressrelease.brainproducts.com/referencing/

Andrey1994:
Hi and thanks for willing to implement it!

I think method signature should look like this: `void reference(BrainFlowArray<double, 2> data, std::vector<int> channels_to_reference, std::vector<int> reference_channels)`

That should be done for all supported languages. 

In terms of implementation first of all you need to average data from reference_channels and after that subtract the averaged data from channels_to_reference

In BrainFlow in general we implement everything in C\C++ and call methods from dyn libs from other languages, but here since implementation is very simple(just average multiple arrays and subtract) and since it requires 2d arrays I think it's ok to just implement it for each language

Andrey1994:
Hi and thanks for working on it!

It should be a static method of DataFilter class for all languages

kkashiva:
Hi @Andrey1994 The reference method implementation in DataFilter class is ready for review for the java_package.
I'll work on similar logic to implement this in other languages (Java is my strongest, so I started there).

I just have 1 doubt on where will this method be called from?

sorenjmadsen:
Anyone done this one in Python? Happy to implement in the dyn lib too

PR #630, opened 2023-05-27
Method to reference channels for Issue #619

kkashiva:
added method signature and implementation comments to DataFilter.java

Creating WIP draft pull request for the issue #619 

Next: complete implementation for java package and then do the same for other language packages.

### DIFF
--- a/java_package/brainflow/src/main/java/brainflow/DataFilter.java
+++ b/java_package/brainflow/src/main/java/brainflow/DataFilter.java
@@ -1041,6 +1041,27 @@ public class DataFilter
     }
 
     /**
+     * (re)reference channels
+     * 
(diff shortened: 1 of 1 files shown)


Andrey1994:
Hi and thanks for working on it.

Java part looks fine.

You can add it to one of example we already have, I do not think that we need a new one.
Downsampling should be ok - https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow/src/main/java/brainflow/examples/Downsampling.java just keep in mind that this example also used as a code samples at BrainFlow Docs, so write some comments there for people who use it as a reference.

Andrey1994:
no activiti in PR, closing

PR #634, opened 2023-06-01
Add NTL Wifi Board

justinmccombs:
As discussed on our call, I believe I have everything included here.


src/board_controller/board_controller.cpp
```diff
@@ -256,6 +256,7 @@ int prepare_session (int board_id, const char *json_brainflow_input_params)
         case BoardIds::GALEA_SERIAL_BOARD_V4:
             board = std::shared_ptr<Board> (new GaleaSerialV4 (params));
             break;
+
```
Andrey1994, line 259: no need for new line

Andrey1994 (changes requested):
Looks good! Just a few minor changes needed.

In addition to comments, you also need to add new board id to all bindings. Like:

https://github.com/brainflow-dev/brainflow/blob/master/python_package/brainflow/board_shim.py#L70

https://github.com/brainflow-dev/brainflow/blob/master/java_package/brainflow/src/main/java/brainflow/BoardIds.java

And the same for remaining programming languages

Andrey1994:
Regarding failed jobs, CI builds it with warnings as errors. To test it locally you can build it with `python tools/build.py --warnings-as-errors`

For failed clang-format job check this https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style

Andrey1994:
@justinmccombs I've fixed the errors, actually it was not working and even was not added to board_controller.cpp. I didnt test these changes since I have no device, but at least it compiles now and looks fine

PR #646, opened 2023-07-20
[WIP] Feature/bind js

GordonPmnt:
This is the very very beginning of JS bind.
Goal is more to show the approach and open PR to contributions.


nodejs_package/brainflow/board-shim.ts
```diff
@@ -0,0 +1,226 @@
...
+    const arch = os.arch();
+    switch (platform) {
+      case 'darwin':
+        return `${this.libPath}/libBoardController.dylib`;
+      case 'win32':
+        return arch === 'x64' ? `${this.libPath}/BoardController.dll` : `${this.libPath}/BoardController32.dll`;
+      case 'linux':
+        return `${this.libPath}/libBoardController.so`;
+      default:
+        throw new Error(`OS ${platform} is not correctly handed by Cortex Desktop App, please contact support.`);
```
GordonPmnt, line 86: Oops. Must change this err message here

nodejs_package/brainflow/brainflow.types.ts
```diff
@@ -0,0 +1,231 @@
...
+  CLASSIFIER_IS_NOT_PREPARED_ERROR = 21,
+  ANOTHER_CLASSIFIER_IS_PREPARED_ERROR = 22,
+  UNSUPPORTED_CLASSIFIER_AND_METRIC_COMBINATION_ERROR = 23,
+}
+
+// --------------------------- //
+//  Interfaces
+// --------------------------- //
+
+export interface IInputParams {
```
Andrey1994, line 214: Could you explain a little bit why it should be an interface?

Andrey1994 (commented):
looks great so far!
I think it makes sense to complete all methods in BoardShim class and move to DataFilter and MLModel after that.

I can help with docs, examples, CI and cmake scripts to copy libs if you need

GordonPmnt:
> looks great so far! I think it makes sense to complete all methods in BoardShim class and move to DataFilter and MLModel after that.
> 
> I can help with docs, examples, CI and cmake scripts to copy libs if you need

Makes sense indeed. Thanks Andrey.
Your help is more than welcome for scipts, CI, cmake... you have way more experience than me on that. I'll focus on JS transcription.

I also might need some examples / hints about the unit testing approach. :)

nodejs_package/brainflow/brainflow.types.ts
GordonPmnt: Yes sure. 

In typescript, it is common practice to use `interface` to define the structure of an object, not only for classes.

Some people prefer using `type` instead. I don't have a strong opinion about that to be honest.

nodejs_package/brainflow/board-shim.ts
```diff
@@ -0,0 +1,232 @@
...
+    const prepared = [0];
+    const res = this.boardController.isPrepared(prepared, this.boardId, this.inputJson);
+    if (res !== BrainFlowExitCodes.STATUS_OK) {
+      throw new Error('Could not check prepared state');
+    }
+    return !!prepared[0];
+  }
+
+  public addStreamer(streamerParams: string, preset = BrainFlowPresets.DEFAULT_PRESET): void {
+    const res = this.boardController.addStreamer(streamerParams, preset, this.boardId, this.inputJson);
```
GordonPmnt, line 119: hey @Andrey1994 
In python I see we're doing this:

```py
if streamer_params is None:
            streamer = None
        else:
            try:
                streamer = streamer_params.encode()
            except BaseException:
                streamer = streamer_params
```

I don't see anything similar in Java. Not sure to understand why we're doing this.

Andrey1994:
moved to another PR since here we have some permissions issue

PR #655, opened 2023-07-28
Update ganglion decompression

philippitts:


src/board_controller/openbci/ganglion.cpp
```diff
@@ -479,6 +396,138 @@ void Ganglion::read_thread ()
     delete[] package;
 }
 
+void Ganglion::decompress_firmware_3 (
+
```
Andrey1994, line 400: strange that clang format does not remote it

src/board_controller/openbci/ganglion.cpp
```diff
@@ -592,7 +641,22 @@ int Ganglion::call_open ()
         }
 
         safe_logger (spdlog::level::info, "search for {}", params.mac_address.c_str ());
-        res = func (const_cast<char *> (params.mac_address.c_str ()));
+
+        struct ConnectionParameters
```
Andrey1994, line 645: I dont like strcut definition like this in cpp code and it seems like its duplicated between ganglion.cpp and bglib, can be moved to a common header I think. Also, compiler may align it randomly somewhere(its unlikely)

src/board_controller/openbci/ganglion_bglib/callbacks.cpp
```diff
@@ -68,8 +71,18 @@ void ble_evt_gap_scan_response (const struct ble_msg_gap_scan_response_evt_t *ms
 
     if (name_found_in_response)
     {
-        // strcasestr is unavailable for windows
-        if (strstr (name, "anglion") != NULL)
+        // Determine the firmware version from the advertised name
+        if (msg->data.data[0] == 20 && msg->data.data[13] == '3')
+        {
+            GanglionLib::firmware = 3;
```
Andrey1994, line 77: I think this entire if block should be under the check for expected name pattern

src/board_controller/openbci/ganglion_bglib/callbacks.cpp
```diff
@@ -68,8 +71,18 @@ void ble_evt_gap_scan_response (const struct ble_msg_gap_scan_response_evt_t *ms
...
+        }
+
+        if (memcmp (msg->sender.addr, (uint8 *)GanglionLib::connect_addr.addr, sizeof (bd_addr)) ==
+            0)
+        {
+            GanglionLib::exit_code = (int)GanglionLib::STATUS_OK;
```
Andrey1994, line 83: why is it added?

src/board_controller/openbci/ganglion_bglib/main.cpp
```diff
@@ -100,7 +107,15 @@ namespace GanglionLib
         }
         exit_code = (int)CustomExitCodes::SYNC_ERROR;
         state = State::OPEN_CALLED;
-        char *mac_addr = (char *)param;
+
+        struct ConnectionParameters
```
Andrey1994, line 111: common definition in header will be better

src/board_controller/openbci/ganglion_native.cpp
```diff
@@ -442,10 +444,9 @@ void GanglionNative::adapter_1_on_scan_found (
 
     if (found)
     {
-        {
-            std::lock_guard<std::mutex> lk (m);
-            ganglion_peripheral = peripheral;
-        }
+        firmware = strncmp (peripheral_identified, "Ganglion 1.3", 12) == 0 ? 3 : 2;
```
Andrey1994, line 447: dont remember right now, but more likely there was a reason why I added extra scope with { and } for lock guard. Maybe I wanted this to be released before calling `notify_one` and maybe its more correct...  lets return it

Andrey1994 (commented):
looks ok, I wrote some comments but mostly because its alway scary to touch some code for ganglion..
there is a set of tests we need to pass before merging it:
- old compression works for raw data for both boards
- new compression works for raw data for both boards
- impedance checking works for old and new compression for board boards
- text messages sent from ganglion(bytes 200-207) work for both boards and both algorithms

src/board_controller/openbci/ganglion.cpp
philippitts: I tried to make it one line, but the clang-format settings move the parameters to the second line. I think it may be because the line is longer than the max line length. Is there a different format you'd like this in?

src/board_controller/openbci/ganglion_bglib/callbacks.cpp
philippitts: The OpenBCI GUI uses the serial_number brainflow parameter to pass the advertised name of the expected device. This is important because if two Ganglions are active, the user will pick one or the other device from the GUI. Brainflow needs to know which device to connect to. This could probably be done via a MAC address, but since this is the way it's currently being done, I haven't changed it.

src/board_controller/openbci/ganglion_bglib/callbacks.cpp
philippitts: See the note about the serial_number parameter below. If a specific device name is passed to brainflow, the firmware must still be determined. Putting this block outside of the name and serial_number checks prevents duplicating the same logic in both locations.

Andrey1994: I mean that its probably more reliable to check fw after validating the name, some data from other devices can be send and first they should be filtered by the name. Here you will set fw value wo such filtering

src/board_controller/openbci/ganglion_bglib/callbacks.cpp
Andrey1994: ok, got it. before we were doing filtering by mac address if I remember correctly and gui passed mac addressses instead names(user still picked the name)

src/board_controller/openbci/ganglion_bglib/callbacks.cpp
Andrey1994: or two ganglions in the same place with two different fw versions, wo check for name you can mess up here

src/board_controller/openbci/ganglion.cpp
philippitts: Oh weird - should be fixed now.

philippitts:
> looks ok, I wrote some comments but mostly because its alway scary to touch some code for ganglion.. there is a set of tests we need to pass before merging it:
> 
>     * old compression works for raw data for both boards
> 
>     * new compression works for raw data for both boards
> 
>     * impedance checking works for old and new compression for board boards
> 
>     * text messages sent from ganglion(bytes 200-207) work for both boards and both algorithms

Test results below. Old compression isn't supported on a device with the new firmware so I don't think we can test both compression types with both boards, but I will show examples of both the new and old firmwares working. Let me know if you had something else in mind.

I'm also not sure what makes sense to test for the text messages sent from Ganglion. The impedance values are sent in bytes 201 - 205. You can query the device registers and get text message responses, but AFAIK there is no way in brainflow to send this command.

This is a 4mV at 4Hz sine wave using a board with firmware 3.0.0 (new)
image (1)

This is a 4mV at 4Hz sine wave using a board with firmware 2.0.1 (old). It looks bad, but this is expected because this was one of the cases where the old compression algorithm exhibited failures. It is similar to results for the same sine wave using the latest brainflow release.
image (2)

This is the impedance check running with the new firmware. I don't have leads on me right now, but the channels do generally respond as expected when I touch the pins. There are no crashes or errors, and the responses are similar on both the old and new firmwares. We will do more comprehensive testing (with leads) after assembling the full tech stack.
image

This is the impedance check running with the old firmware.
image

Andrey1994 (approved):

Andrey1994:
looks ok, I will review and merge PR for freeeg128 and create a new release this weekend

PR #656, opened 2023-07-30
move unpack_from_jar to its own class

Andrey1994:


Andrey1994:
it adds support for non mac devices to #519

PR #662, opened 2023-08-03
Freeeeg128

TedAhmadi:
I fixed the issue we have with a compile error, can you please compile it again?


src/board_controller/freeeeg128/freeeeg128.cpp
```diff
@@ -0,0 +1,232 @@
...
+        delete serial;
+        serial = NULL;
+    }
+    return (int)BrainFlowExitCodes::STATUS_OK;
+}
+
+void FreeEEG128::read_thread ()
+{
+    int res;
+    constexpr int max_size = 200; // random value bigger than package size which is unknown
```
Andrey1994, line 127: @TedAhmadi did you test it?

I dont fully understand how it can work, since here max size is 200 and you set min_package_size to 1 + 128*3 which is bigger

Andrey1994: actually there should be a segfault

Andrey1994:
I pushed some changes to avoid copypaste between freeeg32 and 128, please pull and test them. I suppose there was a bug and more likely it didnt work before and I cannot test it

Andrey1994:
also you need to validate sampling rate, record data for 10 seconds, store in file and confirm that there are around `10 * sampling rate(256)` rows in resulting file

daki90 (approved):

Andrey1994:
no updates, and still not sure if it works or not. That should not damage the operation of freeeg32, so I will merge it, but will not update website with info about freeeg128 support and will not promote it anywhere

PR #663, opened 2023-08-06
Add log message for EOF in PlaybackFileBoard.

stellarpower:
BrainFlow's stream doesn't block, so, the user calling into the session has to use some other method - sleeping, or waiting for user input before terminating the session.

When playing back a file, and not looping, it's possible the file will ned before the user is expecting. At this point, the file reader will sleep, but no message is printed, so the pipeline silently stops sending values.

This simply adds a logging line for when the file playback virtual board has reached the end of file and stops producing data.

### DIFF
--- a/src/board_controller/playback_file_board.cpp
+++ b/src/board_controller/playback_file_board.cpp
@@ -246,7 +246,8 @@ void PlaybackFileBoard::read_thread (int preset, std::string file)
         }
         if ((!loopback) && (res == NULL))
         {
-            if (!reached_end){
+            if (!reached_end)
+            {
                 reached_end = true;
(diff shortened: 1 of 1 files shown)


Andrey1994:
https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style

To fix failed ci job

Also failed job prints the diff

stellarpower:
The instructions there don't say how to run clang-format, so it might be worth adding something that can be pasted into the command-line. I have installed clang-format and given it the same argument as were used in the CI check and it's not happy, it looks that most of the options have been removed, so I just gave up and pasted the patch lines back into the browser. From work I also believe some colleagues have used more recent versions and that could apply changes re casing and identifier names, so may be of use for you.

PR #665, opened 2023-08-13
Feature/bind js

Andrey1994:
recreate #646 from local branch, cannot push to remote branch, #646 will be closed and we will continue from here


retiutut:
Awesome! If boards are able to use native bluetooth from phones, could this be a step forward for using React Native for cross-platform mobile dev + BrainFlow? Or do you just envision this binding unlocking client/server desktop web apps?

Andrey1994:
Its only for desktop app development.

Maybe somehow it will work on android for boards which already supported for android but will not help with ble

PR #672, opened 2023-09-09
Fixed confliciting name

pasotee:
Running on MacOs with M2 chip, I was getting the following error while building due to conflict names:

Screenshot 2023-09-08 at 20 12 14

Since both `size` and `length` are taken I opted for `num`, but feel free to suggest a different name.


Andrey1994:
looks good, strange that CI didnt complain about it even with warnings as errors

PR #675, opened 2023-09-19
Support Ganglion v3 detection on MacOS native BLE

philippitts:
Firmware version detection for Ganglion depends on parsing the advertised name of the device. On MacOS, the device name is always discovered as "Simblee" rather than the name intended to be advertised by the firmware. Since the device name is not determined correctly, the correct Ganglion firmware cannot be determined on MacOS.

This PR solves the above problem by reading the BLE Software Revision String characteristic. This characteristic explicitly provides the correct device firmware version.

### DIFF
--- a/src/board_controller/openbci/ganglion_native.cpp
+++ b/src/board_controller/openbci/ganglion_native.cpp
@@ -8,6 +8,7 @@
 
 #define GANGLION_WRITE_CHAR "2d30c083-f39f-4ce6-923f-3484ea480596"
 #define GANGLION_NOTIFY_CHAR "2d30c082-f39f-4ce6-923f-3484ea480596"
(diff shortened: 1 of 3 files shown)


retiutut:
Are y'all sure this doesn't break other BLE devices? @philippitts @Andrey1994

retiutut:
Reviewed. 👍

src/board_controller/openbci/ganglion_native.cpp
```diff
@@ -444,7 +454,6 @@ void GanglionNative::adapter_1_on_scan_found (
 
     if (found)
     {
-        firmware = strncmp (peripheral_identified, "Ganglion 1.3", 12) == 0 ? 3 : 2;
```
retiutut, line 447: Cool

PR #679, opened 2023-10-13
Add check for os.arch when unpacking MLModel .dlls on WIndows

retiutut:
@Andrey1994  This change will cleanup the console log for all BrainFlow users who work with Java, including OpenBCI Public GUI and Galea GUI.

Currently, an error is thrown every time the MLModel class is instantiated.

### DIFF
--- a/java_package/brainflow/src/main/java/brainflow/MLModel.java
+++ b/java_package/brainflow/src/main/java/brainflow/MLModel.java
@@ -59,7 +59,7 @@ public class MLModel
                     JarHelper.unpack_from_jar ("onnxruntime_arm64.dll");
                     break;
                 default:
-                    throw new RuntimeException ("Unsupported Windows architecture: " + arch);
+                    System.err.println("Unsupported Windows architecture: " + arch);
             }
         } else if (SystemUtils.IS_OS_MAC)
         {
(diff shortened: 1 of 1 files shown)


java_package/brainflow/src/main/java/brainflow/MLModel.java
```diff
@@ -43,10 +43,24 @@ private interface DllInterface extends Library
...
+                    JarHelper.unpack_from_jar ("onnxruntime_x64.dll");
+                    break;
+                case "arm":
+                    JarHelper.unpack_from_jar ("onnxruntime_arm.dll");
+                    break;
+                case "arm64":
+                    JarHelper.unpack_from_jar ("onnxruntime_arm64.dll");
+                    break;
+                default:
+                    throw new RuntimeException ("Unsupported Windows architecture: " + arch);
```
Andrey1994, line 62: lets replace exception by writing to stderr

Andrey1994 (commented):
looks good! lets replace exception by stderr since its an optional dependency and unpacking itself can be optional too

java_package/brainflow/src/main/java/brainflow/MLModel.java
retiutut: Replaced with `System.err.println()`

PR #681, opened 2023-10-19
Update Galea v4 emulator to control sampling rate

retiutut:


Andrey1994:
part about adding new command is fine but I dont get the motivation for debug messages and all these calculations, seems like it doesnt change anything in terms of sampling rate

retiutut:
> part about adding new command is fine but I dont get the motivation for debug messages and all these calculations, seems like it doesnt change anything in terms of sampling rate

It works as expected on the receiving end.

retiutut:
@Andrey1994 @philippitts Can we review and merge? This has many important additions.

Andrey1994:
I've fixed failed alpine test and removed some logic which made it too complicated for simple emulator. The proper tracking and emulation of sampling rate behaviour would require another approach and more refactoring. And after all sampling rate matching is not a requirement for functional high level model, keeped logic for channel settings and notifications for sampling rate commands

PR #686, opened 2023-11-13
add get_firmware_version command to ganglion boards

Andrey1994:


Andrey1994:
@philippitts should do the trick I guess

PR #699, opened 2024-01-12
add explore plus code for 8 and 32 chan

salman2135:


python_package/run_brainflow.py
```diff
@@ -0,0 +1,17 @@
+from brainflow.board_shim import BoardShim, BrainFlowInputParams, BoardIds
+
+BoardShim.enable_dev_board_logger()
+
+
+params = BrainFlowInputParams()
+params.mac_address = "00:13:43:A1:84:EE"
```
Andrey1994, line 7: if you want to use it as an example it makes sense to add mac address as an optional argument, put it in examples/tests folder and rename to smth like explore_plus_8_channels_test.py

Andrey1994 (commented):
looks good

PR #707, opened 2024-02-22
Update Galea default command and log Galea gain values

isaacplotkin:


Andrey1994 (approved):

Andrey1994 (commented):
obviously that will work, but maybe it would make more sense to get them from the board instead internal map which can be wrong in case of hw issues.
But I dont know the intention behind this. In case if you want programmatic access to get info about particular channel I would change that to smth like `get_gains_for_channel_%num%`. Otherwise you will need to parse the string

isaacplotkin:
We were having issues with the GUI and FW gain settings not matching up with the Brainflow settings. We wanted a way to print all Brainflow gains for debugging purposes. I understand it is not the cleanest solution but it works for our use case.

PR #711, opened 2024-04-01
Tiny fixups

fredldotme:
Correctly adds corrections to improve correctness.

Allows building BrainFlow as a submodule.


Andrey1994:
thanks

PR #730, opened 2024-07-01
Brainflow BA_v4 integration

Sarthak031:


Andrey1994:
I think macos job is failed because they've removed macos 11 runners frim github and its for entire repo, this issue with manylinux does not look related to your changes either, so you can ignore it, I fill fix in master branch

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -6,7 +6,22 @@
 #include "timestamp.h"
 
 // common constants
-#define BRAINALIVE_PACKET_SIZE 224
+#define BRAINALIVE_PACKET_SIZE 220
```
Andrey1994, line 9: nitpicking but I think it could be slightly better to use constexpr instead defines because all these params are used only inside single method and can be easily defined there with constexpr

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
 void BrainAlive::read_data (simpleble_uuid_t service, simpleble_uuid_t characteristic,
     uint8_t *data, size_t size, int channel_num)
 {
-    if (size != BRAINALIVE_PACKET_SIZE)
+    static int SOFTWARE_GAIN, HARDWARE_GAIN, REFFRENCE_VOLTAGE;
```
Andrey1994, line 389: there is no reason to make these static afaict, and we dont use capital case for variables

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
...
+    if (size == BRAINALIVE_HANDSHAKING_PACKET_SIZE)
+    {
+        // printf ("inside 6\n");
```
Andrey1994, line 392: remove debugging printf

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
...
+        SOFTWARE_GAIN = data[1];
+        HARDWARE_GAIN = data[2];
+        REFFRENCE_VOLTAGE = (data[3] << 8) | data[4];
+    }
+    else if (size != BRAINALIVE_PACKET_SIZE)
```
Andrey1994, line 397: lets invert the logic, to have == here and final else with error

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
...
-                BRAINALIVE_EEG_SCALE_FACTOR / BRAINALIVE_EEG_GAIN_VALUE;
+            double ba_data[15] = {0};
+            int k = 0;
+            for (int j = i + BRAINALIVE_EEG_DATA_START_INDEX; j < i + BRAINALIVE_EEG_DATA_END_INDEX;
+                 j += 3, k++)
+            {
+                ba_data[k] = (float)(((data[j] << 16 | data[j + 1] << 8 | data[j + 2]) << 8) >> 8) *
+                    ((((float)REFFRENCE_VOLTAGE * 1000) /
+                        (float)(SOFTWARE_GAIN * HARDWARE_GAIN * FSR_Value)));
+                // printf ("%f,", ba_data[k]);
```
Andrey1994, line 415: remove printf

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
...
+            }
+            for (int j = i + BRAINALIVE_ACCLR_MTR_DATA_START_INDEX;
+                 j < i + BRAINALIVE_ACCLR_MTR_DATA_END_INDEX; j += 2, k++)
+            {
+                ba_data[k] = (data[j] << 8) | data[j + 1];
+                if (ba_data[k] > 32767)
+                    ba_data[k] = ba_data[k] - 65535;
+                // printf ("%f,", ba_data[k]);
```
Andrey1994, line 423: and here

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
...
+            }
+            ba_data[14] = data[BRAINALIVE_PACKET_ID_INDEX + i];
+            //  printf ("%f\n", ba_data[14]);
```
Andrey1994, line 426: and here

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +386,45 @@ void BrainAlive::adapter_1_on_scan_found (
...
+    else
     {
-        double eeg_data[9] = {0};
-        for (int j = i + 4, k = 0; j < i + 28; j += 3, k++)
+
+        for (int i = 0; i < (int)size; i += BRAINALIVE_SINGLE_PACKET_SIZE)
         {
-            eeg_data[k] = (((data[j] << 16 | data[j + 1] << 8 | data[j + 2]) << 8) >> 8) *
-                BRAINALIVE_EEG_SCALE_FACTOR / BRAINALIVE_EEG_GAIN_VALUE;
+            double ba_data[15] = {0};
```
Andrey1994, line 407: I dont understand how it works, here the size is 15 and hardcoded, in brainflow_boards.cpp num_rows is 18, and it should match each other, otherwise it will not be aligned in memory properly. I would recommend to allocate this buffer dynamically once somewhere using smth like this https://github.com/brainflow-dev/brainflow/blob/master/src/board_controller/enophone/enophone.cpp#L103 and release it in the end, This way you will avoid misalignments

src/board_controller/brainflow_boards.cpp
```diff
@@ -710,14 +710,15 @@ BrainFlowBoards::BrainFlowBoards()
...
+        {"marker_channel", 17},
         {"package_num_channel", 0},
-        {"num_rows", 17},
+        {"num_rows", 18},
         {"eeg_channels", {1, 2, 3, 4, 5, 6, 7, 8}},
         {"eeg_names", "F7,FT7,T7,CP5,CZ,C6,FC6,F4"},
         {"accel_channels", {9, 10, 11}},
-        {"ppg_channels", {12, 13, 14}}
+        {"gyro_channels", {12, 13, 14}},
+        {"packet_id", {15}}
```
Andrey1994, line 721: we use package_num_channel for packet id, I see no reason to duplicate them. Even more we do not have an API to query for `packet_id`, its only for `package_num_channel` and its unified among boards. Some device specific data can be written to `other_info` channels, but here its clearly should be `package_num_channel`

Andrey1994 (changes requested):
I will fix CI issues in master branch, the main problem here is wrong buffer size and `packet_id` other comments are more like minor refactoring

src/board_controller/brainflow_boards.cpp
```diff
@@ -710,14 +710,14 @@ BrainFlowBoards::BrainFlowBoards()
...
         {"sampling_rate", 250},
-        {"timestamp_channel", 15},
-        {"marker_channel", 16},
-        {"package_num_channel", 0},
+        {"package_num_channel", 14},
+        {"marker_channel", 15},
+        {"timestamp_channel", 16},
         {"num_rows", 17},
-        {"eeg_channels", {1, 2, 3, 4, 5, 6, 7, 8}},
+        {"eeg_channels", {0, 1, 2, 3, 4, 5, 6, 7}},
```
Andrey1994, line 717: why did you remove package_num_channel from position 0? Its there for all boards, and lets keep it consistent

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -1,24 +1,20 @@
...
 
-// common constants
-#define BRAINALIVE_PACKET_SIZE 224
 
 // info about services and chars
 #define START_BYTE 0x0A
 #define STOP_BYTE 0x0D
+#define BRAINALIVE_HANDSHAKING_PACKET_SIZE 6
+
+int software_gain, hardware_gain, reffrence_volatage;
```
Andrey1994, line 13: as far as I understand it these values are device speciffic and we determine their values in BF during handshacking. It makes a lot of sense to convert them to class fields, set to 0 in constructor, read, and use in notifications

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -1,24 +1,20 @@
...
 #include "timestamp.h"
+#include <string.h>
 
-// common constants
-#define BRAINALIVE_PACKET_SIZE 224
 
 // info about services and chars
 #define START_BYTE 0x0A
 #define STOP_BYTE 0x0D
+#define BRAINALIVE_HANDSHAKING_PACKET_SIZE 6
```
Andrey1994, line 11: these alse can be moved to constexpressinos for consistency

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -371,21 +375,36 @@ void BrainAlive::adapter_1_on_scan_found (
...
-
-    for (int i = 0; i < (int)size; i += 32)
-    {
-        double eeg_data[9] = {0};
-        for (int j = i + 4, k = 0; j < i + 28; j += 3, k++)
+        for (int i = 0; i < (int)size; i += brainalive_single_packet_size)
         {
-            eeg_data[k] = (((data[j] << 16 | data[j + 1] << 8 | data[j + 2]) << 8) >> 8) *
-                BRAINALIVE_EEG_SCALE_FACTOR / BRAINALIVE_EEG_GAIN_VALUE;
+            double ba_data[ba_brainflow_package_size] = {0};
```
Andrey1994, line 382: as discussed earlier it probably makes sense to determine it fron board_descr, and to make it future proof you can use it for package_num_channel, eeg_channels, etc. Here is how it looks like: https://github.com/brainflow-dev/brainflow/blob/c9ae582ca10aa2e2a9eb7ea3e30c71c64d40c001/src/board_controller/freeeeg/freeeeg.cpp#L171 https://github.com/brainflow-dev/brainflow/blob/c9ae582ca10aa2e2a9eb7ea3e30c71c64d40c001/src/board_controller/freeeeg/freeeeg.cpp#L145 this way next time when you change brainflow_boards.cpp you will not need to syncup this array here

Andrey1994 (approved):

PR #735, opened 2024-08-01
Added the the get gain command

Sarthak031:


src/board_controller/brainalive/brainalive.cpp
```diff
@@ -190,6 +192,7 @@ int BrainAlive::prepare_session ()
     if ((res == (int)BrainFlowExitCodes::STATUS_OK) && (control_characteristics_found))
     {
         initialized = true;
+        config_board ("0a036007000d");
```
Andrey1994, line 195: Add check for result here

src/board_controller/brainalive/brainalive.cpp
```diff
@@ -302,15 +305,20 @@ int BrainAlive::config_board (std::string config)
...
     }
-    uint8_t command[5];
-    command[0] = 0x0a;
-    command[1] = 0x81; // it is hardcoded for now only
-    command[2] = 0x00;
-    command[3] = 0x00;
-    command[4] = 0x0d;
-    safe_logger (spdlog::level::trace, config[2]);
+    // Calculate the number of bytes
+    size_t numBytes = config.length () / 2; // Each byte is represented by 2 hex characters
```
Andrey1994, line 309: There is snake case with _ for variables everywhere in brainflow, lets keep it consistent

Andrey1994: Applies to all variables in this block

Andrey1994:
I also dont get why do you need to config board multiple times and what these configs mean, but that board speciffic part is up to your device to determine

src/board_controller/brainalive/brainalive.cpp
Andrey1994: So it should be res =... and below that initialized = true

Sarthak031: done

src/board_controller/brainalive/brainalive.cpp
Sarthak031: done

Sarthak031:
> I also dont get why you need to config board multiple times and what these configs mean, but that board speciffic part is up to your device to determine

The first config is for start data streaming, the second config is used for stop data streaming, and the third is used to get the configuration of the device before data streaming.

Andrey1994 (approved):

PR #737, opened 2024-08-14
Fix interpretation of IMU data for Galea beta

MusaMahmood:
Issue:

- IMU data was being interpreted incorrectly: byte order was wrong, and scale factor was wrong. 
- Assuming ±4g for accelerometer, ±500° for gyroscope and ±1300μT (x,y-axis), ±2500μT (z-axis) for magnetometer

### DIFF
--- a/src/board_controller/openbci/galea_v4.cpp
+++ b/src/board_controller/openbci/galea_v4.cpp
@@ -466,9 +466,12 @@ void GaleaV4::read_thread ()
                 // aux, 5 times smaller sampling rate
                 if (((int)b[0 + offset]) % 5 == 0)
                 {
-                    double accel_scale = (double)(0.002 / (pow (2, 4)));
-                    double gyro_scale = (double)(0.002 / (pow (2, 4)));         // to be confirmed
-                    double magnetometer_scale = (double)(0.002 / (pow (2, 4))); // to be confirmed
+                    double accel_scale = (double)(4.0 / static_cast<double> (pow (2, 16) - 1));
+                    double gyro_scale = (double)(4000.0 / static_cast<double> (pow (2, 16) - 1));
+                    double magnetometer_scale_xy =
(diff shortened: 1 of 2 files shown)


Andrey1994:
https://brainflow.readthedocs.io/en/stable/BrainFlowDev.html#code-style to fix failed  job

Andrey1994 (commented):
I dont understand why byte order for eeg and accel is different, it does not make sense as for me and looks like a bug in fw or very strange design decision, and dont get why it is not like in cyton too.

If you are confident that its correct, I will merge it but I would highly recommend to double check fw and make byte order uniform for eeg and aux data

MusaMahmood:
> I dont understand why byte order for eeg and accel is different, it does not make sense as for me and looks like a bug in fw or very strange design decision, and dont get why it is not like in cyton too.
> 
> If you are confident that its correct, I will merge it but I would highly recommend to double check fw and make byte order uniform for eeg and aux data

I checked accelerometer and gyroscope, they're definitely correct - before it was all noisy. 
It depends IC to IC. For BMI160 and BMM150 it looks like it is LSB first. ADS1299 is MSB first. 

I need to make another fix re: scaling and this should be ready to merge

MusaMahmood:
Ok I just realized magnetometer [BMM150](https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmm150-ds001.pdf), x and y axis is 13-bit and z axis is 15-bit twos-complement, which means `cast_16bit_to_int32` will not interpret correctly. Very annoying.

Andrey1994:
Ok, got it. Maybe it makes sense to unify and simplify it in fw(you can easily swap bytes there), but up to you. You also probably want to do it in serial board tp keep them consistent

Andrey1994 (approved):

PR #751, opened 2024-12-23
Added impedance mode for ANT neuro devices

bdieudonne:


Andrey1994:
Hi, thanks, I will review tomorrow

docs/SupportedBoards.rst
```diff
@@ -1026,12 +1026,20 @@ Ant Neuro has many devices and all of them are supported by BrainFlow:
...
 - :code:`ANT_NEURO_EE_511_BOARD`
 
-Initialization Example:
+Initialization and EEG reading example:
 
 .. code-block:: python
 
     params = BrainFlowInputParams()
-    board = BoardShim(BoardIds.ANT_NEURO_EE_410_BOARD, params)
+    board = BoardShim(BoardIds.ANT_NEURO_EE_410_BOARD, params)  # 8 channel amplifier
```
Andrey1994, line 1034: that should not be here, its meant to be just a part to init the device
device specific code samples can be pushed into python_package/examples/tests
You can create an example with impedance checking and add a link for it to the docs

Andrey1994 (commented):
there is `get_resistance_channels` methods in brainflow api which is supposed to be used to mark impedance values.
ANT neuro devices have quite a lot of channels so I dont think that we should extend the whole package with 2 times more channels, but we can add resistance_channels into `brainflow_boards.cpp` and make them the same as exg channels. simplified example:

```
eeg_channels: {1,2,3,4...}
resistance_channels: {1,2,3,4...}
```

Then in user api it will be possible to write smth like:
```
resistance_channels = BoardShim.get_resistance_channels(board_id)
impedance_data = data[resistance_channels]
```

bdieudonne:
Thanks for reviewing! I pushed some updates based on your comments. Some remarks:

- The listing of eeg/emg/resistance/... channels manually feels a bit error prone. E.g., when you accidentally set the board to AntNeuroEE410 instead of AntNeuroEE411, I lost all my EEG data (because that one has no eeg_channels defined) -- while the ANT sdk itself could easily work device agnostic.
- I put the sampling counter at the end instead of at the beginning of the data package array (I saw the same is done for the gtec Unicorn), to avoid that corresponding channels would have different indices in EEG mode versus in impedance mode.
- I also removed sampling_rate from the brainflow_boards.cpp, as it is not fixed. Example to explain why:

```
board = BoardShim(BoardIds.ANT_NEURO_EE_410_BOARD, params)
board.prepare_session()
board.config_board('sampling_rate:500')
print(board.get_sampling_rate(board.board_id))
```
--> this would yield 2000, while I'd expect 2000

Andrey1994:
>I also removed sampling_rate from the brainflow_boards.cpp, as it is not fixed. Example to explain why:

that should be restored, its known that sampling rate returned by BoardShim can be wrong if its manually overridden with config_board, it comes from the fact that its static info but there must be a default

python_package/examples/tests/eego_impedances_and_eeg.py
```diff
@@ -0,0 +1,29 @@
...
+
+if __name__ == '__main__':
+    params = BrainFlowInputParams()
+    board = BoardShim(BoardIds.ANT_NEURO_EE_411_BOARD, params)  # 8 channel amplifier
+    board.prepare_session()
+
+    # Get impedance data
+    board.config_board('impedance_mode:1')
+    board.start_stream()
+    for i in range(5):
```
Andrey1994, line 13: dont see reasons for the loop here, just wait for 5 seconds and print will do the same

python_package/examples/tests/eego_impedances_and_eeg.py
```diff
@@ -0,0 +1,29 @@
...
+        time.sleep(1)
+        data = board.get_board_data()  # get all data and remove it from internal buffer
+        print(f'{data.shape[0]} channels x {data.shape[1]} samples')
+    board.stop_stream()
+    
+    # Get EEG data
+    board.config_board('impedance_mode:0')
+    board.start_stream()
+    for i in range(3):
```
Andrey1994, line 22: the same here

src/board_controller/ant_neuro/ant_neuro.cpp
```diff
@@ -61,14 +61,7 @@ AntNeuroBoard::AntNeuroBoard (int board_id, struct BrainFlowInputParams params)
     fact = NULL;
     amp = NULL;
     stream = NULL;
-    try
-    {
-        sampling_rate = board_descr["default"]["sampling_rate"];
```
Andrey1994, line 66: need to restore, its used as a default

src/board_controller/ant_neuro/ant_neuro.cpp
```diff
@@ -267,11 +284,27 @@ void AntNeuroBoard::read_thread ()
...
                 }
                 package[board_descr["default"]["timestamp_channel"].get<int> ()] = get_timestamp ();
                 push_package (package);
             }
             std::this_thread::sleep_for (std::chrono::milliseconds (1));
+            if (impedance_mode)
+            {
+                // some more sleep; twice every second should be more than enough
+                // if left out, it yields impedances at around 64 Hz
+                std::this_thread::sleep_for (std::chrono::milliseconds (500));
```
Andrey1994, line 306: is it correct? I dont rememeber how ant sdk works, but it looks like it might lead to the same speed of collecting data and slower speed of reading it from the buffer. So in the end it will result to overflow, lost packages and delay
Also, not sure why I've added sleep above, looks strange for me now also

src/board_controller/brainflow_boards.cpp
```diff
@@ -485,153 +485,161 @@ BrainFlowBoards::BrainFlowBoards()
     brainflow_boards_json["boards"]["24"]["default"] =
     {
         {"name", "AntNeuroEE410"},
-        {"sampling_rate", 2000},
```
Andrey1994, line 488: that should be restored

src/board_controller/brainflow_boards.cpp
```diff
@@ -485,153 +485,161 @@ BrainFlowBoards::BrainFlowBoards()
...
         {"timestamp_channel", 10},
         {"marker_channel", 11},
-        {"package_num_channel", 0},
+        {"package_num_channel", 9},
```
Andrey1994, line 490: package num is the first one for all boards and makes sense to keep it consistent

Andrey1994 (changes requested):

python_package/examples/tests/eego_impedances_and_eeg.py
bdieudonne: It's just a simple example that could be used as a starting point for some processing in blocks of 1 second, that easily generalizes to longer streams. The buffer should be emptied now and then to save memory, right? (Is there any guideline on how often the buffer should be emptied?)

src/board_controller/ant_neuro/ant_neuro.cpp
bdieudonne: From the ANT sdk manual: "Although it is called a stream, it is actually a state. Calling GetData() on an impedance stream returns the latest known impedance state." So there's no internal buffer for impedances

I guess the 1ms sleep is to give the hardware/cpu some rest (and/or allow other threads to do stuff)?

src/board_controller/brainflow_boards.cpp
bdieudonne: Note that ANT EEG streams have N+2 channels (N EEG + triggers + sample counter) and impedance streams have N+2 channels as well (M=N EEG + reference electrode + ground electrode). To have them both fit in the brainflow channels (now: N+2 + marker + time stamp), I saw four "reasonable" options:

1. Keep EEG stream as before --> EEG and impedance channels don't match, and there's no (room for) package num for impedances
    - EEG (as before):
        - chan 0: package num (= sample counter)
        - chan 1-N: EEG channels
        - chan N+1: triggers
        - chan N+2: time stamp channel
        - chan N+3: marker channel
    - impedances:
        - chan 0-N-1: EEG channels
        - chan N: reference channel 
        - chan N+1: ground channel
        - chan N+2: time stamp channel
        - chan N+3: marker channel
2. Set package num at N+1 for the EEG stream --> similar to gtec Unicorn integration, **agrees exactly with the default channel numbering of ANT SDK**, and **EEG and impedance channels match** (also no package num for impedances)
    - EEG:
        - chan 0-N-1: EEG channels
        - chan N: package num (= sample counter)
        - chan N+1: triggers
        - chan N+2: time stamp channel
        - chan N+3: marker channel
    - impedances:
        - chan 0-N-1: EEG channels
        - chan N: reference channel 
        - chan N+1: ground channel
        - chan N+2: time stamp channel
        - chan N+3: marker channel
3. Add an extra channel (N+5 channels) --> everything fits as well, but needs extra channel. Channels don't match exactly between EEG and impedances.
    - EEG:
        - chan 0: package num (= sample counter)
        - chan 1-N: EEG channels
        - chan N+1: triggers
        - chan N+2: empty
        - chan N+3: time stamp channel
        - chan N+4: marker channel
    - impedances:
        - chan 0: empty/manual package num
        - chan 1-N: EEG channels
        - chan N+1: reference channel 
        - chan N+2: ground channel
        - chan N+3: time stamp channel
        - chan N+4: marker channel
4. Add two extra channel (N+6 channels) --> everything fits as well, every channel has its own dedicated index, but needs 2 extra channels
    - EEG:
        - chan 0: package num (= sample counter)
        - chan 1-N: EEG channels
        - chan N+1: empty (~reference)
        - chan N+2: empty (~ground)
        - chan N+3: triggers
        - chan N+4: time stamp channel
        - chan N+5: marker channel
    - impedances:
        - chan 0: empty/manual package num
        - chan 1-N: EEG channels
        - chan N+1: reference channel 
        - chan N+2: ground channel
        - chan N+3: empty (~triggers)
        - chan N+4: time stamp channel
        - chan N+5: marker channel

I think option 1 is a no-go as it is confusing to have the EEG channels not match; I choose option 2 as it agrees best with ANT's sdk. Let me know what you think

src/board_controller/ant_neuro/ant_neuro.cpp
Andrey1994: good, thanks!

python_package/examples/tests/eego_impedances_and_eeg.py
Andrey1994: you can set the size of the buffer in start_stream method, there is a default also, its smth like 3600 * 250 and its a ring buffer

src/board_controller/brainflow_boards.cpp
Andrey1994: I like option 4 with manual package num for impedance, its closer to what is implemented in other boards.

It also will be a breaking change since brainflow has playbackfileboard and read_file methods and all these 4 options will lead to the need to update recorded files. In option 4 it can be made with less harm, package number is at index 0 as before and eeg channels are at indexes 1-N as before, so for example this code will just work with older files:

```
data = DataFilter.read_file("old_ant.data")
eeg_channels = BoardShim.get_eeg_channels(board_id)
eeg_data = data[eeg_channels]
```

We technically can keep timestamps and markers working with old files if we put 

- chan N+1: reference channel
- chan N+2: ground channel
- chan N+3: empty (~triggers)

After timestamp and marker channels and dont change locations for timestamp and marker. But it will look a little bit strange, its common to put timestamps and markers in the end. So lets update it exactly like you propose in option 4

Andrey1994: - chan N+1: reference channel
- chan N+2: ground channel
- chan N+3: empty (~triggers)

And

- chan N+1: empty (~reference)
- chan N+2: empty (~ground)
- chan N+3: triggers

Should go to other_channels, right? Update in docs will be needed to reflect that

bdieudonne: I don't know how other_channels is defined. Would you consider ground and reference as "other channels", if they are already listed in resistance_channels?

bdieudonne: updated according to option 4 (hope i didnt make any typos)

bdieudonne: (it is still a breaking change, as the trigger channel, timestamp channel and marker channel have moved, as you noted)

Andrey1994 (approved):

PR #755, opened 2025-01-07
Update galea_v4 config_board method to receive acknowledgements while streaming

retiutut:


Andrey1994 (commented):
its not gonna work, there is another thread reading from the same socket as a result acks you are hoping to receive here will be received in another thread reading data from the board and vice versa, data you should read in the data acquisition thread will be acquired in config_board method

Andrey1994:
Proper way to handle it would be to create multiple sockets instead

Andrey1994:
will close this one, feel free to create a new one and I can also help implementing it if needed

PR #756, opened 2025-02-23
Fix PyQT Error

cheongpark:
Fixed PyQt and PyQtGraph compatibility issues:

1. QtGui.QApplication -> QtWidgets.QApplication
   **Error**: AttributeError: module 'pyqtgraph.Qt.QtGui' has no attribute 'QApplication'
   **Fix**: Updated import to use QtWidgets which is the correct module for QApplication in newer Qt versions

2. pg.GraphicsWindow -> pg.GraphicsLayoutWidget
   **Error**: AttributeError: module 'pyqtgraph' has no attribute 'GraphicsWindow'
   **Fix**: GraphicsWindow was deprecated and replaced with GraphicsLayoutWidget in newer PyQtGraph versions

3. Added "show=True" parameter
   **Fix**: Explicitly set window visibility when creating GraphicsLayoutWidget

### DIFF
--- a/python_package/examples/enophone/enophone_streaming.py
+++ b/python_package/examples/enophone/enophone_streaming.py
@@ -3,7 +3,7 @@ import logging
 
 import pyqtgraph as pg
 from brainflow.board_shim import BoardShim, BrainFlowInputParams, BoardIds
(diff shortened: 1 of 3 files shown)


Andrey1994:
thanks

PR #760, opened 2025-03-04
Add classify_power_features for BCI state classification

BurhanxGodhra:
Added a function to classify brain states (rest/focus) using SVM on band powers (alpha 8-13 Hz, beta 13-30 Hz). Tested with synthetic data in Colab—predicts based on power ratios. First PR—feedback welcome! Part of my 30-day BCI learning sprint.

### DIFF
--- a/python_package/brainflow/classify_power.py
+++ b/python_package/brainflow/classify_power.py
@@ -1,0 +1,65 @@
+
+!pip install --upgrade brainflow scikit-learn
+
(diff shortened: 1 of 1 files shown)


Andrey1994:
thanks, but its not an extension of brainflow api, just an example and I dont see much value adding it into the brainflow repo keeping in mind that we already have classification examples which use builtin brainflow methods

PR #764, opened 2025-04-30
ANT Neuro SDK functions

bdieudonne:
Added ANT SDK info as custom config_board function `"get_info"`:

```python
board.config_board("get_info")
```
would return (for example):
```json
{
    "type": "EE411",
    "firmware_version": 12345,
    "serial_number": "424242",
    "sampling_rates": [500,512,1000,1024,2000,2048],
    "reference_ranges": [1,0.75,0.15],
    "bipolar_ranges": [1,0.75,0.15],
    "power_state": {
        "is_powered": 1,
        "is_charging": 0,
        "charging_level": -1
    }
}
```

### DIFF
--- a/src/board_controller/ant_neuro/ant_neuro.cpp
+++ b/src/board_controller/ant_neuro/ant_neuro.cpp
@@ -11,8 +11,6 @@
 #include <unistd.h>
 #endif
 
-#include <sstream>
-
 #include <algorithm>
 #include <chrono>
 
@@ -23,6 +21,7 @@
 #include "eemagine/sdk/wrapper.h"
 
 using namespace eemagine::sdk;
+using json = nlohmann::json;
 
 
 AntNeuroBoard::AntNeuroBoard (int board_id, struct BrainFlowInputParams params)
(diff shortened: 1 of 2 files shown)


src/board_controller/ant_neuro/ant_neuro.cpp
```diff
@@ -471,8 +474,73 @@ int AntNeuroBoard::config_board (std::string config, std::string &response)
             return (int)BrainFlowExitCodes::INVALID_ARGUMENTS_ERROR;
         }
     }
+    else if (config == get_info) // return stringified JSON with info from ANT board
+    {
+        std::ostringstream oss;
```
Andrey1994, line 479: do not use stringstream, brainflow imports nlohman json already and you can just constract a json object. It will look cleaner

Andrey1994 (changes requested):
as an idea looks nice, just need to swap to json on cpp side instead stringstream

src/board_controller/ant_neuro/ant_neuro.cpp
bdieudonne: good point, updated! looks cleaner indeed :)

Andrey1994 (approved):

PR #778, opened 2025-08-01
added option to specify serial number for ANT device

### LINKED ISSUE #775: Brainflow does not find the right ANT amplifier when multiple amplifiers are connected to a pc
**Describe the bug**
Brainflow does not find the right ANT amplifier when multiple amplifiers are connected to a pc

**To Reproduce**
Connect two different ANT amplifiers to your computer (e.g., one AntNeuroEE411 and one AntNeuroEE511). Then try:
```python
    board_id = BoardIds.ANT_NEURO_EE_411_BOARD  # or BoardIds.ANT_NEURO_EE_511_BOARD
    board = BoardShim(board_id, params)
    board.prepare_session()
```
One of them will fail with the following error (or the other way around):
`[board_logger] [error] No devices found, Board name AntNeuroEE411 does not match board name of connected device AntNeuroEE511`

**Info**
 - Board Id: ANT_NEURO_EE_XXX_BOARD
 - OS running on your PC: any
 - Programming Language and its version: any
 - Architecture(x86, x64, ARM, etc): any
 - Brainflow version: 0.5.18

**Suggested solution**
1. Update the function `getAmplifier` with optional arguments:
    - amplifier type: the function will return the first connected amplifier with given type
    - (optional) amplifier serial number: the function will return the exact amplifier that is asked for
 2. (optional) Update `prepare_session` to handle params.serial_number

bdieudonne:
Closes #775

### DIFF
--- a/src/board_controller/ant_neuro/ant_neuro.cpp
+++ b/src/board_controller/ant_neuro/ant_neuro.cpp
@@ -118,7 +118,7 @@ int AntNeuroBoard::prepare_session ()
                 (!expected_serial_number.empty () && current_serial != expected_serial_number))
             {
                 safe_logger (spdlog::level::info, 
-                    "Found non-matching amplifier (board name {}, serial {})", current_board_name, 
+                    "Found non-matching amplifier (board name {}, serial {})", current_board_name,
                     current_serial);
                 delete current_amp;
                 continue;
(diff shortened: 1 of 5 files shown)


src/board_controller/ant_neuro/ant_neuro.cpp
```diff
@@ -96,7 +96,8 @@ int AntNeuroBoard::prepare_session ()
         safe_logger (spdlog::level::info, "eego sdk version is: {}.{}.{}.{}",
             fact->getVersion ().major, fact->getVersion ().minor, fact->getVersion ().micro,
             fact->getVersion ().build);
-        amp = fact->getAmplifier ();
+        amp = fact->getAmplifier (board_descr["default"]["name"], params.serial_number);
```
Andrey1994, line 99: there is `getAmplifiers` in the ant neuro sdk, I think it would be better to use that instead and iterate over the list and check for serial number and name if needed. This way will also lead to better loging of erros like wrong name, wrong serial number, etc.

PR #781, opened 2025-09-26
updated ANT neuro wrapper

bdieudonne:
- Impedances are not read for bipolar channels, so they should not be in the list of resistance_channels.
- To be sure that emg_channels are not filled with garbage from in EEG mode, the whole package is filled with zeros before filling with data.


Andrey1994:
looks good to me, will create a new release next week

PR #801, opened 2026-04-10
iOS support for C++ API (no Swift bindings yet)

kcoul:
Currently under test. In draft until all changes are verified.


Andrey1994:
cmake changes look ok to me but I cannot test them myself

Andrey1994:
could you add verification of it into github actions pipelines?

Andrey1994:
anu updates?

kcoul:
Thanks for reminder, I'll add the GitHub action pipeline and it should be ready to merge. 
We could do the Swift bindings in a separate PR in the future.

kcoul:
This should also be ready for review now @Andrey1994 - again, @calebkuitenbrouwer has added the GitHub Action for iOS C++ build. Let us know if anything still needs work.

.github/workflows/run_ios.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    steps:
+    - name: Clone Repository
+      uses: actions/checkout@v4
+
+    - name: Configure (iOS Simulator arm64)
+      run: |
+        cmake -B build \
+          -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains/ios.toolchain.cmake \
+          -DPLATFORM=SIMULATORARM64 \
```
Andrey1994, line 17: its simulator only, for real devices it should be OS64, is the intent of this file just to check that its built and not to distribute compiled libs?

Andrey1994 (commented):
my main question about this change what is the plan going forward?
as is it builds libs but for appstore/ios/etc should all of it be packed as XCFramework? should it be one XCFramework per compiled library?

kcoul:
> my main question about this change what is the plan going forward? as is it builds libs but for appstore/ios/etc should all of it be packed as XCFramework? should it be one XCFramework per compiled library?

I think we should do the XCFramework as a subsequent PR - we might be able to make two artifacts - one for unmanaged C++ API and the other with Swift bindings (did someone contribute a PR for that? I'll take another look...)

.github/workflows/run_ios.yml
kcoul: Yes we can get this updated, I think @calebkuitenbrouwer did not have a real hardware target but might be able to switch this over now. We could add that change to a packaging PR in the near future.

Andrey1994:
@kcoul there is https://github.com/brainflow-dev/brainflow/pull/834 my main question for this PR is packaging, not the code itself, but would appreciate if you can review this
cc @tsvvladimir

Issue #604, opened 2023-01-28
Windows build error (MLModule.dll copy step)


kcoul:
**Describe the bug**
Build of BrainFlow target, in Release mode, fails with:

Error copying file (if different) from "C:/Users/username/repos/brainflow/compiled/Release/MLModule.dll" to "C:/Users/username/repos/brainflow/python_package/brainflow/lib/MLModule.dll".

Upon inspecting the source folder, MLModule.dll is located at "C:/Users/username/repos/brainflow/compiled/MLModule.dll", not in a "Release" subfolder. This explains the problem.

**To Reproduce**
Build in CLion for Windows.

Andrey1994:
Hi, we do not support clion oficially... and as far as I understand clion uses mingw build tools to compile projects on windows by default. And mingw also is not currently supported(#255).
More likely it will start working with clion as soon as #255 is fixed but I cannot provide the ETA for it.

If you do not like visual studio I think you can download VS community edition and use it only for building code and continue using clion for development - https://www.jetbrains.com/help/clion/quick-tutorial-on-configuring-clion-on-windows.html#MSVC

kcoul:
Good to know! For the record, CLion can in fact use Visual Studio for a generator (it uses Ninja generator by default). By using both the Visual Studio toolchain *and* the Visual Studio generator, the post-build copy step is successful.

PR #802, opened 2026-04-17
Fix MLModule post-build copy path for MSVC generators

baris-talar:
Fixes an inconsistency in `src/ml/build.cmake` where MSVC POST_BUILD steps assume a config subdirectory (`compiled/$<CONFIG>/...`) while output is set to `compiled/`.

Replaces hardcoded source paths with target-based generator expressions:
- `$<TARGET_FILE:${ML_MODULE_NAME}>`
- `$<TARGET_LINKER_FILE:${ML_MODULE_NAME}>`

This improves compatibility with single-config generators (e.g. Ninja in CLion) where artifacts may not be placed in config-specific subdirectories.

Related to #255 and #604.


Andrey1994:
looks good to me, but I believe board controller and data handler use the same pattern and should be updated the same way

baris-talar:
Thanks!

Before I update those, could you clarify the expected scope so I don’t over-apply the change:

- Should I apply the same $<TARGET_FILE> / $<TARGET_LINKER_FILE> replacement to all MSVC POST_BUILD copy commands in board_controller and data_handler?
- Or only to targets where output directories are set to a non-config-specific path like `compiled`?

If possible, I’d really appreciate as much detail as you can provide (e.g. specific files, targets, or patterns to update), so I can apply the fix correctly and keep the patch consistent.

Andrey1994:
I think everything what uses pattern like `${CMAKE_CURRENT_SOURCE_DIR}/compiled/$<CONFIG>/NAME_HERE` can be replaced with it.

Also with this is there a real need for:

```
if (msvc)
....
else
```

The idea to add this if else was to handle $CONFIG and different paths as a result for different OSes

Andrey1994:
```
if (UNIX AND NOT ANDROID)
    add_custom_command (TARGET ${ML_MODULE_NAME} POST_BUILD
        COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${CMAKE_CURRENT_SOURCE_DIR}/compiled/${ML_MODULE_COMPILED_NAME}" "${CMAKE_CURRENT_SOURCE_DIR}/nodejs_package/brainflow/lib/${ML_MODULE_COMPILED_NAME}"
        COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${CMAKE_CURRENT_SOURCE_DIR}/compiled/${ML_MODULE_COMPILED_NAME}" "${CMAKE_CURRENT_SOURCE_DIR}/python_package/brainflow/lib/${ML_MODULE_COMPILED_NAME}"
        COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${CMAKE_CURRENT_SOURCE_DIR}/compiled/${ML_MODULE_COMPILED_NAME}" "${CMAKE_CURRENT_SOURCE_DIR}/julia_package/brainflow/lib/${ML_MODULE_COMPILED_NAME}"
```

In the file you have changed this part probably can be removed now? Just make sure we dont copy it for ANDROID

baris-talar:
Thanks, I’ve updated the same pattern in ml, board controller, and data handler, while keeping Android logic and external dependency copies unchanged.

Andrey1994:
@codex

PR #832, opened 2026-04-29
Add Shimmer3 board support

VigneshRajan-AMRC:
##### _Updated PR message to reflect the new commits_

Add native C++ driver for Shimmer3 (Bluetooth SPP)

- New board: BoardIds.SHIMMER3_BOARD (id 68)
- Channels: 3-axis low-noise accelerometer, gyroscope, magnetometer, GSR, ExG, battery, and temperature
- Data: raw ADC values streamed with host-side timestamps
- Runtime config via config_board(): sampling_rate:<Hz> and sensors:<hex>
- Scope: Shimmer3 (hardware version 3) only; prepare_session() rejects Shimmer3R (hardware version 10)

Tested on Windows with a Shimmer3 over Bluetooth SPP via the Python binding


Andrey1994:
Hi, it looks good at the first sight. I will check more details in the next couple of days.

One concern I have - gpl license for these files, I am not 100% sure it doesnt conflict with MIT license in brainflow.

Also, I can update docs, website, etc, do you have images you want to use or any wishes what to highlight there?

VigneshRajan-AMRC:
Hi @Andrey1994 , thanks for looking into my PR.

I just looked into the GPL3 license terms and you were right to be concerned, it can't be merged into any other license. I assumed GPL3 was compatible with any open-source license, but I was wrong. 

Anyways, I will spend the next few days reworking the Shimmer3 Cpp API from scratch using one of the ShimmerResearch published API repos. They all have 3-Clause BSD licenses and they can be merged into an MIT project as long as their llcense terms and owner are mentioned in the code header and probably in the readme too. So, that wouldn't affect the Brainflow MIT license.

Until then, could you please keep the 68 board ID blocked for the Shimmer3? You could possibly accept the changes to the boardids and boardshims in all the lang-packs, as they aren't affected by any licensing mismatch.

And thanks for offering to update the docs and website. I mainly created and tested this for the Shimmer3 GSR+ device ~~but the API should be compatible with other Shimmer3/3R units as well~~. Here's [the user-guide](https://shimmersensing.com/wp-content/docs/support/documentation/GSR_User_Guide_rev1.13.pdf) for the Shimmer3 GSR+ with relevant information for the docs. The pictures of the device are better from [this site](https://imotions.com/products/hardware/shimmer3-gsr/).

Andrey1994:
hi! do you need any help with that?

Andrey1994:
any updates?

VigneshRajan-AMRC:
Hi @Andrey1994, thanks for offering to help. I couldn't focus on this work past few days but have started working on this again and aiming to finish up sometime this week.

VigneshRajan-AMRC:
Hi @Andrey1994 , I have now reworked the Shimmer3 C++ API files from scratch. I have also made some refinements along the way. I updated the original PR message to reflect the current PR.

The latest changes adds support only for the Shimmer 3 variant and not the 3R. That board is a newer revision and requires separate API. I have updated the previous comment where I mentioned otherwise, just for clarity. I am considering adding support to the 3R later, once I have the board.

The images and information required for updating the docs are still valid from my previous comment.

> Here's [the user-guide](https://shimmersensing.com/wp-content/docs/support/documentation/GSR_User_Guide_rev1.13.pdf) for the Shimmer3 GSR+ with relevant information for the docs. The pictures of the device are better from [this site](https://imotions.com/products/hardware/shimmer3-gsr/).

src/board_controller/shimmer/shimmer3.cpp
```diff
@@ -0,0 +1,694 @@
...
+
+// Block until len bytes are read or we give up.
+int Shimmer3::read_exact (uint8_t *buf, int len)
+{
+    if (serial_port == nullptr)
+        return (int)BrainFlowExitCodes::BOARD_NOT_CREATED_ERROR;
+
+    int got = 0;
+    int idle = 0;
+    const int max_idle = 2000; // ~2s of empty reads before timing out
```
Andrey1994: its not 2 seconds, its 2000 seconds in fact.
There is read timeout cconfigured a line 323 which is 1000ms, so 2000 attempts here leads to 2000 seconds for which theread calling brainflow api will be blocked

src/board_controller/shimmer/shimmer3.cpp
```diff
@@ -0,0 +1,694 @@
...
+    {
+        int r = cmd_inquiry ();
+        if (r == (int)BrainFlowExitCodes::STATUS_OK)
+            resp = "OK";
+        return r;
+    };
+
+    if (config.rfind ("sampling_rate:", 0) == 0)
+    {
+        double hz = std::stod (config.substr (14));
```
Andrey1994: that may throw an exception which will not be catched, and it will crash the calling process, especially since its called from other bindings.
Make sure it returns brainflow exit code in case of error instead

src/utils/inc/brainflow_constants.h
```diff
@@ -96,6 +96,7 @@ enum class BoardIds : int
     IRONBCI_32_BOARD = 65,
     NEUROPAWN_KNIGHT_BOARD_IMU = 66,
     MUSE_S_ATHENA_BOARD = 67,
+    SHIMMER3_BOARD = 68,
     // use it to iterate
     FIRST = PLAYBACK_FILE_BOARD,
     LAST = MUSE_S_ATHENA_BOARD
```
Andrey1994: last should be updated to SHIMMER3_BOARD

src/board_controller/shimmer/shimmer3.cpp
```diff
@@ -0,0 +1,694 @@
...
+
+    serial_port = Serial::create (port_name.c_str (), this);
+    if (serial_port->open_serial_port () < 0)
+    {
+        safe_logger (spdlog::level::err, "failed to open {}", port_name);
+        delete serial_port;
+        serial_port = nullptr;
+        return (int)BrainFlowExitCodes::UNABLE_TO_OPEN_PORT_ERROR;
+    }
+    serial_port->set_serial_port_settings (1000, false);
```
Andrey1994: not checking for returrn code

src/board_controller/shimmer/inc/shimmer3.h
```diff
@@ -0,0 +1,77 @@
...
+{
+public:
+    Shimmer3 (struct BrainFlowInputParams params);
+    ~Shimmer3 ();
+
+    int prepare_session () override;
+    int start_stream (int buffer_size, const char *streamer_params) override;
+    int stop_stream () override;
+    int release_session () override;
+    int config_board (std::string config, std::string &response) override;
```
Andrey1994, line 27: in brainflow api new method was added - get_board_sampling_rate, you can use this one

Andrey1994: especially since you allow overriding sampling rates via config_board method, it would be good to have it, check https://github.com/brainflow-dev/brainflow/pull/805

Andrey1994 (changes requested):
wrote some comments but overall looks good

src/board_controller/shimmer/shimmer3.cpp
```diff
@@ -0,0 +1,694 @@
...
+            put_axis ("gyro_channels", gyro_axis++ % 3);
+            break;
+
+        case Signal::MAG_X:
+        case Signal::MAG_Y:
+        case Signal::MAG_Z:
+            put_axis ("magnetometer_channels", mag_axis++ % 3);
+            break;
+
+        case Signal::GSR:
```
Andrey1994, line 812: for eda we need calculated values, not raw values, quick tip from codex which I havent checked -
```
For Shimmer3 GSR, the 16-bit raw value is not directly EDA. It contains a 12-bit ADC value plus, in auto-range mode, range bits.

  Use this:

  int raw16 = ...;                  // little-endian uint16 from packet
  int adc = raw16 & 0x0FFF;         // lower 12 bits
  int range = configured_gsr_range; // 0..3 fixed, 4 auto

  if (range == 4)
      range = (raw16 >> 14) & 0x03; // top two bits encode actual range

  Then convert to conductance:

  struct Coeff { double p1; double p2; };

  static constexpr Coeff coeffs[4] = {
      {0.0373,    -24.9915}, // range 0
      {0.0054,    -3.5194},  // range 1
      {0.0015,    -1.0163},  // range 2
      {0.0004558, -0.3014}   // range 3
  };

  double eda_us = coeffs[range].p1 * adc + coeffs[range].p2; // microSiemens

  Equivalent resistance formula from Shimmer’s API is:

  double resistance_kohm = 1000.0 / eda_us;

  For BrainFlow, I would put eda_us into eda_channels, not raw ADC and not kOhms. Existing Galea code stores EDA conductance in microSiemens at /d:/workspace/brainflow/src/
  board_controller/openbci/galea.cpp:484. If you still want raw/debug data, put raw ADC or raw16 into other_channels.

  Also handle invalid values:

  if (range < 0 || range > 3)
      return INVALID;

  if (eda_us <= 0.0)
      eda_us = 0.0; // or NaN, depending on desired BrainFlow behavior
```

src/board_controller/shimmer/shimmer3.cpp
```diff
@@ -0,0 +1,694 @@
...
+
+    uint8_t hdr[8];
+    res = read_exact (hdr, 8);
+    if (res != (int)BrainFlowExitCodes::STATUS_OK)
+        return res;
+
+    uint16_t div = static_cast<uint16_t> (hdr[0] | (hdr[1] << 8));
+    sampling_rate = divider_to_hz (div);
+    uint8_t num_channels = hdr[6];
+    uint8_t buffer_size = hdr[7];
```
Andrey1994: its read but ignored, from the docs found when this value is more than 1 data packet can containt multiple timestamps and channel samples

src/board_controller/shimmer/shimmer3.cpp
```diff
@@ -0,0 +1,694 @@
...
+            // "device timestamp" style row exists, otherwise ignored here.
+            break;
+
+        case Signal::ACCEL_LN_X:
+        case Signal::ACCEL_LN_Y:
+        case Signal::ACCEL_LN_Z:
+        case Signal::ACCEL_WR_X:
+        case Signal::ACCEL_WR_Y:
+        case Signal::ACCEL_WR_Z:
+            put_axis ("accel_channels", accel_axis++ % 3);
```
Andrey1994: need to make sure accel x goes to accel_chnnels[0], y to accel_channels[1] and z goes to accel_channels[2]
currently its not guaranteed

Andrey1994 (commented):
a few more comments

VigneshRajan-AMRC:
Hi @Andrey1994, I just saw the changes you've requested as I've been busy and away last month. I will look into these in the next couple of weeks.

VigneshRajan-AMRC:
This update addresses the Shimmer3 review feedback and improves runtime configuration and Bluetooth packet handling.

Changes include:
- Catch and translate invalid 'config_board()' inputs.
- Validate sampling-rate dividers and 24-bit sensor masks.
- Check serial-port configuration failures.
- Report the active device sampling rate through 'get_board_sampling_rate()'.
- Parse the inquiry buffer-size field and decode every buffered sample.
- Convert GSR ADC values to conductance in microsiemens (μS) for fixed and automatic ranges.
- Route accelerometer, gyroscope, and magnetometer axes by explicit signal ID.
- Reject configurations whose physical IMU sources would overwrite the same BrainFlow rows.
- Reject unsupported inquiry fields instead of risking packet de-synchronisation.
- Add cancellable serial reads and atomic stream-thread control.
- Synchronise 'start_stream()' with receipt of the first valid sample.
- Correct Shimmer3 sensor masks, signal IDs, and packet field formats.

The buffering implementation follows the Shimmer BtStream inquiry and data packet format, where the inquiry buffer-size value determines how many samples are combined into one Bluetooth packet.

src/board_controller/shimmer/shimmer3.cpp
VigneshRajan-AMRC: Fixed. Good catch.

src/utils/inc/brainflow_constants.h
VigneshRajan-AMRC: Fixed.

src/board_controller/shimmer/inc/shimmer3.h
VigneshRajan-AMRC: Fixed. I added get_board_sampling_rate(int preset). It returns the active inquiry-derived sampling rate for the default preset, including changes made through config_board(), and delegates other presets to the base implementation.

src/board_controller/shimmer/shimmer3.cpp
VigneshRajan-AMRC: Fixed. The result of set_serial_port_settings() is now checked against SerialExitCodes::OK. Configuration failures are logged, the port is closed and released, and a public BrainFlow error is returned instead of continuing with a partially configured connection.

src/board_controller/shimmer/shimmer3.cpp
VigneshRajan-AMRC: Fixed. Sampling-rate and sensor-mask parsing are now guarded by exception handling and return INVALID_ARGUMENTS_ERROR for invalid input. The code also verifies full-string consumption, finite and representable rates, the 24-bit mask limit, supported sensor bits, and conflicting sensor selections.

src/board_controller/shimmer/shimmer3.cpp
VigneshRajan-AMRC: Fixed. The driver now extracts the configured GSR range from the inquiry configuration. Fixed-range mode uses that configured range, while automatic mode uses the range bits included in each GSR sample. The lower 12-bit ADC value is converted to conductance in microsiemens before being written to eda_channels.

src/board_controller/shimmer/shimmer3.cpp
VigneshRajan-AMRC: Fixed. Accelerometer, gyroscope, and magnetometer values are now routed by explicit signal ID, so X/Y/Z consistently map to channel indices 0/1/2 regardless of inquiry order. MPU accelerometer and magnetometer signals use the same explicit routing when selected.

src/board_controller/shimmer/shimmer3.cpp
VigneshRajan-AMRC: Fixed. The inquiry buffer-size field is stored as samples_per_packet. The streaming thread reads the complete packet_data_size * samples_per_packet payload and separately decodes and pushes each timestamp/channel set, so no buffered sample bytes remain unread in the serial stream. The protocol defines buffer size as the number of samples sent together in one packet.

Andrey1994 (approved):

PR #844, opened 2026-07-31
Improve Muse PPG example usability

### LINKED ISSUE #843: Improve usability of python_package/examples/tests/muse_ppg.py
Use case

I recently built BrainFlow from source with BLE support and tested python_package/examples/tests/muse_ppg.py using a physical Muse 2 on macOS.

The example worked successfully and produced PPG data, oxygen saturation, and heart-rate estimates. While using it, I noticed a few usability issues that make it harder to use as a diagnostic/example script:

* The recording duration is hard-coded to 200 seconds.
* There are no standard BrainFlow CLI arguments for selecting a device (--mac-address, --serial-number, --timeout, --other-info).
* The script prints the entire raw data matrix before the final results.
* The oxygen saturation and heart-rate values are printed without labels.
* The session cleanup is not protected with try/finally, so an interruption may skip cleanup.

I think improving these areas would make the example easier for new users to work with while preserving its current functionality.

Describe the solution you’d like

I’d like to submit a small PR that:

* Preserves the existing default behavior (including the 200-second recording duration).
* Adds optional CLI arguments following the pattern used by other Python examples.
* Uses try/finally to ensure stop_stream() and release_session() are always called.
* Labels the oxygen saturation and heart-rate outputs.
* Optionally avoids printing the full raw data matrix by default while still allowing users to inspect it if desired.

I would not change:

* BrainFlow APIs
* Muse preset handling (p50)
* BLE/device behavior
* The oxygen saturation or heart-rate algorithms

Describe alternatives you’ve considered

The current example is functional as-is, so another option is to leave it unchanged.

I also considered submitting the changes directly as a pull request, but I wanted to check first that this type of example cleanup would be welcome before spending time implementing it.

zgoldwyn:
Summary

This PR improves the usability of the Muse 2 PPG example without changing its core functionality.

Closes #843 

Changes

* Added optional CLI arguments:
    * --duration
    * --mac-address
    * --serial-number
    * --timeout
    * --other-info
* Preserved the existing default recording duration of 200 seconds
* Added try/finally cleanup to ensure stop_stream() and release_session() are called appropriately
* Added labels to the SpO₂ and heart-rate outputs
* Added a user-friendly check that reports when too few PPG samples have been collected for heart-rate calculation instead of raising an exception
* Fixed a typo in a comment

Validation

* Built BrainFlow locally with BLE support
* Tested with a physical Muse 2 on macOS
* Verified successful connection, streaming, and data collection
* Verified CLI argument parsing
* Verified cleanup on successful runs
* Verified short recordings now produce a clear message when insufficient data is available for heart-rate calculation
* Ran:
    * python -m py_compile
    * python muse_ppg.py --help
    * invalid duration validation
    * git diff --check

Notes

This PR intentionally does not modify:

* BrainFlow public APIs
* Muse BLE behavior
* Muse preset handling (p50)
* SpO₂ or heart-rate algorithms
* Core C++ code

The goal is to improve the usability and robustness of the example while preserving its existing functionality.

### DIFF
--- a/python_package/examples/tests/muse_ppg.py
+++ b/python_package/examples/tests/muse_ppg.py
@@ -1,34 +1,70 @@
+import argparse
 import time
 
 from brainflow.board_shim import BoardShim, BrainFlowInputParams, BoardIds, BrainFlowPresets
 from brainflow.data_filter import DataFilter
 
+
 def main():
     BoardShim.enable_dev_board_logger()
     DataFilter.enable_dev_data_logger()
 
+    heart_rate_fft_size = 2048
+
+    parser = argparse.ArgumentParser(
+        description='Record Muse 2 PPG data and calculate SpO2 and heart rate.')
+    parser.add_argument(
+        '--duration', type=float,
+        help='recording duration in seconds; longer recordings may be needed for heart-rate calculation (default: 200)',
+        required=False, default=200)
+    parser.add_argument('--mac-address', type=str, help='mac address or macOS device UUID', required=False, default='')
+    parser.add_argument('--serial-number', type=str, help='serial number or device identifier', required=False,
+                        default='')
+    parser.add_argument('--timeout', type=int, help='timeout for device discovery or connection', required=False,
+                        default=0)
+    parser.add_argument('--other-info', type=str, help='other info', required=False, default='')
+    args = parser.parse_args()
+    if args.duration <= 0:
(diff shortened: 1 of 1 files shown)


Andrey1994:
Hi, the change itself looks good and its how brainflow should be used most of the time. But these examples are meant to be single screen simple snippets, I dont want to overload them with no brainflow specific parts on purpose

PR #858, opened 2026-08-23
Add moving average option for ML classifiers (fixes #640)

### LINKED ISSUE #640: Add moving average option to ml classifiers scores
For real time data its a common approach to use moving average for eeg signal score, we can add it directly into brainflow as another classifier

samerzumot:
- Implement `MovingAverageClassifier` in C++ core supporting custom scores and built-in metric models
- Add `MOVING_AVERAGE_CLASSIFIER` enum across Python, Java, C#, TypeScript, Rust, Swift, Julia, and MATLAB bindings
- Add automated test `moving_average_classifier.py`


Andrey1994 (changes requested):
Hi, I was thinking about adding moving average option to already existing classifiers instead of adding a new classifier type.
To enable it or not other_info field can be used.
In terms of implementation will need to think how to make it generic and uniform for all classifier types

samerzumot:
Hi @Andrey1994,

Thanks for the feedback! I've updated the implementation according to your suggestion:

1. **Removed the Standalone Classifier**
2. **Generic Implementation in `BaseClassifier`**:
   - This makes moving average support completely generic and uniform across all classifier types—supporting both scalar outputs (`Mindfulness`, `Restfulness`) and multi-channel vector outputs (`ONNX`, `DynLib`).
   - Third-party `.dll`/`.so` plugin symbol resolution (`"prepare"`, `"predict"`, `"release"`) in `DynLibClassifier` remains 100% unchanged.
3. **`other_info` Parsing**:
   - Accepts JSON (`{"window_len": 4}`, `{"moving_average": true}`, `{"period": 5}`), key-value (`moving_average=5`, `window_len=4`), or simple integer string (`"3"`).
   - Defaults to `DEFAULT_MOVING_AVERAGE_WINDOW = 5` when a flag without a window length is provided.
   - Moving average is only activated if recognized keys are found; arbitrary strings or unrelated JSON configs for custom plugins leave moving average disabled to prevent unintended collisions.
4. **Memory Bounded**: The sliding window history is explicitly trimmed upon every prediction cycle.
5. **Tests**: Updated `moving_average_classifier.py` with comprehensive tests covering baseline raw predictions, window smoothing, restfulness score inversion, default fallbacks, key-value syntax, negative tests, and multi-element vector output via `DynLibClassifier`.

Please take a look when you get a chance!

python_package/examples/tests/moving_average_classifier.py
```diff
@@ -0,0 +1,215 @@
...
+                assert np.allclose(vec_out3, [20.0, 40.0, 60.0])
+
+                vec_model.release()
+                print("Vector output moving average test passed successfully!")
+
+    print("All moving average classifier tests passed successfully!")
+
+
+if __name__ == '__main__':
+    test_moving_average_classifier()
```
Andrey1994: add this one to CI tests here - https://github.com/brainflow-dev/brainflow/blob/master/.github/workflows/run_windows.yml 
also smth like pytest and different methods would look better than one big function

src/ml/base_classifier.cpp
```diff
@@ -60,3 +67,213 @@ int BaseClassifier::set_log_file (const char *log_file)
...
+    {
+        return;
+    }
+
+    std::string info = params.other_info;
+    try
+    {
+        if (info.find ('{') != std::string::npos)
+        {
+            json j = json::parse (info);
```
Andrey1994: this try catch block is too big, lets simplify it and stick either to json format either to =(I prefer json)

Andrey1994 (changes requested):
I like how it looks now, thanks!

about formating and parsing I wrote some comments, plus there are logical errors also, Copypaste from codex:

```
[P2] Existing plugin configuration can silently enable smoothing.
Bare integers and generic keys such as period are interpreted globally. I tested a DYN_LIB plugin using other_info="3" to select channel 3: expected outputs [10, 20] became [10, 15]. Require an explicit smoothing option so existing plugin settings retain their meaning. Code
[P2] Explicitly disabling smoothing does not reliably work.
{"moving_average":false,"window_len":3} still averages predictions because parsing window_len overwrites the flag. Key-value syntax is also order-dependent: moving_average=false;window_len=3 enables smoothing, while reversing the fields disables it. Explicit false should take precedence. Code
[P2] Window size 1 behaves differently across configuration formats.
moving_average=1 selects the default five-sample window, whereas {"moving_average":1} and window_len=1 pass predictions through. Numeric window values should have consistent semantics; reserve true for enabling the default window. Code
```

samerzumot:
Hi @Andrey1994,

Thanks for the feedback and Codex report! I committed the following changes:

- **JSON-only parsing (`base_classifier.cpp`)**: Removed key-value/bare-string parsing completely. Sticking strictly to JSON with `!j.is_object()` guard ignores bare integers/strings (e.g. `other_info="3"` for plugin channels), keeping existing plugin settings intact. Dropped `period`.
- **Precedence for `false`**: `{"moving_average": false}` returns immediately so explicit `false` strictly takes precedence over `window_len`.
- **Consistent window 1**: Numeric `1` (`{"window_len": 1}`, `{"moving_average": 1}`) passes predictions through unchanged; boolean `true` enables default window 5.
- **Refactored tests (`moving_average_classifier.py`)**: Split into modular pytest-style `test_*` functions with dedicated tests for explicit disable, window 1 pass-through, and non-JSON string rejection.
- **CI**: Added the test to `.github/workflows/run_windows.yml` and `run_unix.yml`.

python_package/brainflow/ml_model.py
```diff
@@ -38,7 +38,7 @@ class BrainFlowModelParams(object):
     :type classifier: int
     :param file: file to load model
     :type file: str
-    :param other_info: additional information
+    :param other_info: additional information or configuration (e.g. moving average smoothing via `{"window_len": 5}` or `moving_average=5`). Note: moving average options in other_info are handled uniformly across all classifiers.
```
Andrey1994, line 41: [P2] This documents `moving_average=5` as supported, but the revised parser accepts JSON objects only, and `test_non_json_string_ignored` explicitly expects this exact syntax to be ignored. Please remove the key-value example (or implement it, which would reintroduce the compatibility concern); JSON-only is preferable.

Andrey1994 (changes requested):
Two issues remain in the latest revision:

- The public Python documentation still advertises `moving_average=5`, although parsing is now intentionally JSON-only and the test explicitly verifies that this string is ignored.
- The current Clang Format workflow is failing in `base_classifier.cpp` and `mindfulness_classifier.cpp`.

The earlier configuration-collision and precedence issues are otherwise addressed.

samerzumot:
Fixed. Thanks @Andrey1994

Andrey1994:
thanks!

